Issue #14, opened 2015-10-20
Formatting a file in place


avh4:
`elm-format MyFile.elm` should format the file in place.

Janiczek:
I'll try working on this for now (and then #15, #16, if time allows), will submit a pull request when done.

avh4:
Sounds great, thanks!

avh4:
Fixed in #34

Issue #16, opened 2015-10-20
Formatting all elm files in a directory


avh4:
`elm-format src` (`src` is a directory) should format all `*.elm` files found (recursively) in the given directory.

Using `--output FILE` with a directory should exit with an error message.

The following prompt should be given (unless `--yes` is specified):

```
This will overwrite the following 4 files to use Elm’s preferred style:

  src/TodoMVC.elm
  src/TodoMVC/Model.elm
  src/TodoMVC/Update.elm
  src/TodoMVC/View.elm

This cannot be undone! Make sure to back up these files before proceeding.

Are you sure you want to overwrite these files with formatted versions? (y/N)
```

Janiczek:
What should be the behaviour when you have eg. 3 `.elm` files and the first one has compiler errors in it?
- print something, exitFailure (leaving the remaining files as are)
- print something, and continue parsing the rest of the files?

avh4:
I'd say do the simplest option that seems technically correct to you, and open a new issue to discuss whether we want to handle it in a more complicated way.

avh4:
Fixed in #41

Issue #15, opened 2015-10-20
Interactive warning when formatting a file in place


avh4:
`elm-format path/MyFile.elm` should produce the following prompt:

```
This will overwrite the following files to use Elm’s preferred style:

  path/MyFile.elm

This cannot be undone! Make sure to back up these files before proceeding.

Are you sure you want to overwrite these files with formatted versions? (y/N)
```

`elm-format --yes path/MyFile.elm` or `elm-format path/MyFile.elm --yes` should overwrite the file without prompt.

avh4:
As a further note, the prompts should feel consistent with prompts used in `elm-package`, code for reference: https://github.com/elm-lang/elm-package/blob/master/src/Install.hs

Janiczek:
Oh, amazing - I had this implemented somehow but this is much nicer and
more polished (Cmd.yesOrNo etc.). Will update to that style before creating
a pull request.

PR #34, opened 2015-10-21
Formatting a file in place (#14)

Janiczek:
I have not done Haskell in a long time, so there are probably some code style changes you'll want me to do. (I'm particularly unsure about the `when` ... and maybe the definition of `isOutputEmpty` could use a `$`?

And it would probably make more sense for the `output` option to be `Maybe FilePath` instead of a `FilePath` that can be `""` ... if you want me to refactor it that way, I can try.

Otherwise the functionality is as promised:
```bash
./elm-format MyFile.elm # formats it in place, but doesn't empty it if there's a parse error
./elm-format MyFile.elm --output MyFile2.elm # creates MyFile2.elm, writes the result if successful or leaves empty if parse error
```

(No interactive warning yet.)

### DIFF
--- a/src/Flags.hs
+++ b/src/Flags.hs
@@ -9,7 +9,7 @@ import qualified Text.PrettyPrint.ANSI.Leijen as PP
 
 
 data Config = Config
-    { _file :: FilePath
+    { _input :: FilePath
     , _output :: FilePath
     }
 
@@ -33,7 +33,7 @@ parse =
 flags :: Opt.Parser Config
 flags =
     Config
-      <$> file
+      <$> input
       <*> output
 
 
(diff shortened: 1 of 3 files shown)


avh4:
Looks pretty good.  Yes, let's use `Maybe FilePath`.  Also, can you try to add something to `tests/run_tests.sh` to cover the new functionality.  (Or if you can think of a good way to test this in haskell, that'd be okay too.)

Janiczek:
OK this should be up to speed and ready to be merged. Then #15 could be (I hope) trivial.

Janiczek:
Oh wait, the tests. Forgot about them.

Janiczek:
Is it ready for merging now?

avh4:
Looks great, I should be able to merge it tonight (PST)

Janiczek:
I'll be able to work on #15 and #16 then. (CEST :) )

avh4:
I made a couple minor changes when merging: https://github.com/avh4/elm-format/compare/06e2cda5413449e576f65d31d9c400f35179c7e7...c27324e746f8e5641f13bba473515f661ceec339

Great tests, btw.  Thanks!

PR #40, opened 2015-10-23
Overwrite prompt (#15)

### LINKED ISSUE #15: Interactive warning when formatting a file in place

Janiczek:
Should close #15

### DIFF
--- a/src/CommandLine/Helpers.hs
+++ b/src/CommandLine/Helpers.hs
@@ -1,0 +1,13 @@
+module CommandLine.Helpers where
+
+import System.IO
+
+yesOrNo :: IO Bool
+yesOrNo =
+  do  hFlush stdout
+      input <- getLine
+      case input of
+        "y" -> return True
+        "n" -> return False
+        _   -> do putStr "Must type 'y' for yes or 'n' for no: "
+                  yesOrNo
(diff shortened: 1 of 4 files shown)


Janiczek:
I was looking at the Travis CI. It says something about redundant imports - it perhaps sees some of these functions by default, but my developer machine needs to have them imported explicitly.

About the type signature for showErrors - I got a bit stuck on that so I just let the type inference do its work. If you have a tip for what the signature could be, I can add it.

Travis:
```
      showErrors :: forall (t :: * -> *).
                    Foldable t =>
                    t (RA.Located Syntax.Error) -> IO ()
```

I suppose the forall and typeclass are unimportant, but I don't know what should the last row then be.

Apanatshka:
try:

```haskell
showErrors :: [RA.Located Syntax.Error] -> IO ()
```

Janiczek:
Yes that's it. Thanks!

PR #41, opened 2015-10-26
Format directories

Janiczek:
Relates to issue #16 (hopefully closes it).

The functional changes:
- we check for existence of the INPUT file/directory and exit appropriately
- we check for existence of any .elm files in the INPUT directory and exit appropriately
- we search the directory recursively for .elm files and always overwrite them in place (asking beforehand), exiting on any failure during the formatting.
- because of no indicator of which file did an error happen in, we write something along the lines of "Processing file xyz.elm" before we start working on that file.

The questions are:

1. (#42) should we always try to format as many files as possible or should we exit on the first failure?
2. (#43) should we allow for usage `elm-format DIR --output DIR2`, copying the filenames one for one there?
3. should I try to somehow inject the file-with-error path into the now useless "<location>" placeholder inside the error report, or is it good as it is ("Processing file xyz.elm")? This seems like a nontrivial task, modifying the parser to allow for an extra argument etc.

The code probably needs some refactoring (at least extracting the error messages into a different module?) but I need a bit of a break from it so that I can see it "with a new set of eyes" :) Suggestions welcome! (And some Haskell style.)

### DIFF
--- a/elm-format.cabal
+++ b/elm-format.cabal
@@ -52,6 +52,7 @@ Executable elm-format
         directory >= 1.0 && < 2.0,
         edit-distance >= 0.2 && < 0.3,
         filepath >= 1 && < 2.0,
+        filemanip >= 0.1 && < 1.0,
         indents >= 0.3 && < 0.4,
         language-glsl >= 0.0.2 && < 0.3,
         mtl >= 2.2 && < 3,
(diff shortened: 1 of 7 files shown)


avh4:
I'm going to merge what's done already.

For question (3), where we want to show the filename is in place of "<location>" currently in Main.hs#printError

For extracting strings, I'd like to do it in a way that prepares for internationalization, probably with this approach: https://wiki.haskell.org/Internationalization_of_Haskell_programs_using_Haskell_data_types

PR #55, opened 2015-11-20
Replace hardcoded strings with typed messages

### LINKED ISSUE #45: Extract messages in Main.hs to localizable file
See https://wiki.haskell.org/Internationalization_of_Haskell_programs_using_Haskell_data_types

Janiczek:
Fixes #45.

Is it appropriate to do the same thing to Flags.hs? If yes, which parts?


avh4:
Great!  Thanks for figuring this out.  I think `Flags.hs` can wait for now.

Issue #52, opened 2015-11-08
Get rid of deprecated stuff in src/Box.hs


avh4:
Everything below `DEPRECATED BELOW THIS LINE` should go away, and a new `render :: Box -> String` function should be written.  The dependency on `Text.PrettyPrint.Boxes` should also be removed from `elm-format.cabal`.

avh4:
Finished in #58

PR #58, opened 2015-11-24
New render function

Janiczek:
Will be closing #52.

But not yet - it's still WIP - I have to understand the **Stack** type in `src/Box.hs`. Maybe the whole model of **Line**s (simple enough) and **Box**es.

@avh4, could you explain how do **MustBreak** and **Stack** (and possibly **Margin** and **Empty**, if you think you'll need them later) behave, what do they represent, etc.?

I can then, along with the functional changes, write up some docs explaining this (and lowering the entry barrier) for future contributors.

----

- [x] renderLine (**Text** text)
- [x] renderLine (**Row** lines)
- [x] renderLine **Space**
- [x] renderLine **Tab**
- [x] render (**MustBreak** line)
- [x] render (**Stack** line moreLines)
- [x] BoxTest


avh4:
Sure,  so the `Box` type represents something to display, and will be at least one line high (it cannot be empty).  Boxes either of two classifications: either they can appear in the middle of a line (only the case if the Box itself only has one line), or it must appear by itself (either it is multiple lines, or it is "MustBreak", meaning it is a single line, but nothing can follow it on the line).  "MustBreak" is only used for `--` comments.

src/Box.hs
```diff
@@ -51,6 +49,7 @@ len l =
         Text s -> length s
         Row ls -> sum $ map len ls
         Space -> 1
+        -- Tab -> ??? 0 ? TODO
```
avh4: I'm not quite sure what to do with this either.. Tab has a different width depending on where it is in the line, but also this function is used to figure out how much space is used before a tab, and is never actually called with `Tab`.

src/Box.hs
```diff
@@ -199,71 +198,18 @@ elmGroup innerSpaces left sep right forceMultiline children =
...
+    text
+renderLine (Row ls) =
+    concat $ map renderLine ls
+renderLine Space =
+    " "
+renderLine Tab =
+    "\t"
+
+render :: Box -> String
+-- render (Stack line children) =
```
avh4: The values for Stack are `Stack firstLine moreLines`, and they all just go on top of each other.

src/Box.hs
avh4: Just realized what to do about this:  `len :: Line -> Int` should be replaced with `lineLength :: Int -> Line -> Int` where the first int is the start column, then it makes sense for all cases including `Tab`

Janiczek:
About the possibility of empty list: what if we used something like [Data.List.NonEmpty](http://hackage.haskell.org/package/semigroups-0.18.0.1/docs/Data-List-NonEmpty.html), catching it in the type system? Is that too much of a hassle?

----

Also I have some problems running the tests:
```bash
elm-format $ tests/run-tests.sh 
Re-configuring with test suites enabled. If this fails, please run configure manually.
Resolving dependencies...
Configuring elm-format-0.1...
cabal: At least the following dependencies are missing:
HUnit >=1.1 && <2,
QuickCheck ==2.*,
test-framework >0.8 && <0.9,
test-framework-hunit ==0.3.*,
test-framework-quickcheck2 ==0.3.*,
union-find ==0.2.*,
wl-pprint -any
```
Do you have any idea on how to help it? (Probably some `cabal` command, but I don't know.)

Janiczek:
Right now my implementation of `renderLine Tab` is wrong (see [tests](https://travis-ci.org/avh4/elm-format/builds/94277052#L429)):
```diff
-    a\n    b\n
+\ta\n\tb\n
```
```diff
-    a\n
+ \ta\n
```

So, should the Tab's length be `4 - (startColumn mod 4)` so that this holds?
```haskell
tabLength 0 = 4
tabLength 1 = 3
tabLength 2 = 2
tabLength 3 = 1
tabLength 4 = 4
tabLength 5 = 3
... etc.
```

Janiczek:
@avh4 What do you think?

avh4:
Thanks!   I didn't look at detail at the whole thing, but it looks simple enough, all the tests pass.  Seems good!  Also, this change improves the performance by a huge amount for some reason.

Issue #67, opened 2015-11-27
Format multiple files in a single run


avh4:
`elm-format` should accept more than a single file or directory to be given on the command line.

If multiple files are given, specifying `--output` should be an error.

eeue56:
+1, it was the first thing that I tried. If I get time I'll look into this

avh4:
FYI, currently you can specify a directory, and all `*.elm` files in that directory will be formatted.

But a contribution would be great.  `src/Main.hs`, and `src/Flags.hs` are the place to start.

eeue56:
Ah, perfect! That saves me some time :)

eeue56:
Okay, so I have a working implementation [here](https://github.com/eeue56/elm-format/tree/handle-multiple-files)

I have some questions that you might be able to answer:

- What's the expected behaviour when one of the files passed in isn't found or isn't a valid Elm path?
- Should the `output` flag be ignored when multiple paths are given? I assume so, since that's what happens when you give it a `dir` 
- Should it recurse into directories by default? Eg, if you do `elm-format src` should it then go and format those in `src/Http/` too?

avh4:
If `--output` is specified when there are multiple files given, we should exit with an error message.

Recursion should use `findAllElmFiles` in https://github.com/avh4/elm-format/blob/master/src/ElmFormat/Filesystem.hs which does recurse into subdirectories.

If some files are errors and others are fine, we haven't completely decided what to do.  But specifying multiple files should be consistent with specifying a directory that contains multiple files.

eeue56:
I opened a PR #69 to fix this - let me know what you think!

avh4:
Fixed in #69

PR #69, opened 2015-11-29
Handle multiple files

eeue56:
outfoxes https://github.com/avh4/elm-format/issues/67 as defined in the comments there

# What to test

- Passing directories works as expected
- using globbing (`*.elm`, `src/*`) works as expected
- Passing a single file with an output works as expected
- Passing an output when giving multiple input files fails as expected

### DIFF
--- a/src/Flags.hs
+++ b/src/Flags.hs
@@ -9,7 +9,7 @@ import qualified Text.PrettyPrint.ANSI.Leijen as PP
 
 
 data Config = Config
-    { _input :: FilePath
+    { _input :: [FilePath]
     , _output :: Maybe FilePath
     , _yes :: Bool
     }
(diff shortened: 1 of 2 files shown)


avh4:
Thanks!  I will get to this in the next few days.

If you have a chance to update the `tests/run-tests.sh` script to add some checks for this, that would be great.  (That script is in need of some clean up, though.)  If you can figure out a way to write some Haskell tests for this, that would be even better.

eeue56:
Sure, I'll add some tests. Should be easy to write Haskell tests for everything apart from `main` itself, though could probably do `main` too.

avh4:
Merged, thanks a lot!

There is a new issue for adding haskell tests for Main if you can help with that #72

Issue #66, opened 2015-11-27
Format stdin


avh4:
*As a build tool author,
I want `elm-format` to be able to read from stdin and write to stdout
So that my build tool can format Elm code without first having to write it to a file*

--

Running `elm-format --stdin` should read stdin instead of a file, and write the result to stdout.

Review the usage docs to make sure they are consistent with this and mention that this can be done.

avh4:
To support #65

eeue56:
How do you see end-of-input working? Usually on UNIX it's `Ctrl-d`. Would you expect the same here?

avh4:
I would expect whatever haskell mechanism is used for reading stdin should provide a way of determining end-of-input.  See http://ubuntuforums.org/showthread.php?t=1795636 for one example.  (I don't know haskell that well to know if the example there is the best way to read stdin or not.)

See also http://unix.stackexchange.com/a/110248 for background on how EOF normally works.

eeue56:
Yup - just confirming that you wanted it to work with actual EOF rather than some other way of inputting text. I should get time this weekend to take a stab at this and moving some of the `Main` tests over.

avh4:
Cool!

I don't know what other options there would be, but I'd be curious to hear about any good ones.

eeue56:
This would the best way to do it I think - just wanted to check the spec before I made it. 

I don't think there's really any other nice way of doing it, other than maybe opening an editor session from their settings (like git does for commits). Then they could paste it, it would be saved to a temp file and formatted, then giving them the option to save it back to a real file or copy it to clipboard. But I don't really think that's necessary here!

avh4:
Ah, yes, the reason for adding this is to support IDE plugins and other build tools that will want to run elm-format without needing to have the source code saved in a file.  (See #65)  So this definitely needs to be non-interactive.

avh4:
Fixed in #82

PR #82, opened 2015-12-12
Read from stdin

eeue56:
Outfoxes https://github.com/avh4/elm-format/issues/66 

I also changed the spec slightly -

- If `elm-format -` run, then read from stdin 
- If `elm-format --stdin` run, then read from stdin
- If `--output` is set, then save to that output file

Ideally, I think I'd like it so that there was only a `--stdin` since it's possible to accidentally do `-` and be trapped in a stdin reader. I added the `--stdin` option, so that it self documents the ability to automatically it print it out when the file is run with no args or `--help` is set.

The only thing I haven't done is update the docs as part of this PR - I wanted to check with @avh4 whether we want to keep `elm-format -` as a thing or to ditch it first.

### DIFF
--- a/src/CommandLine/Helpers.hs
+++ b/src/CommandLine/Helpers.hs
@@ -1,6 +1,17 @@
 module CommandLine.Helpers where
 
 import System.IO
+import System.Exit (exitFailure, exitSuccess)
+import Messages.Types (Message(..))
+import Messages.Strings (renderMessage)
+
+import qualified Reporting.Annotation as RA
+import qualified Reporting.Report as Report
+import qualified Reporting.Error.Syntax as Syntax
+
+
+r :: Message -> String
+r = renderMessage
 
 yesOrNo :: IO Bool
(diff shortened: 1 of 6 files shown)


avh4:
Okay, I'm sold on removing `elm-format -` in favor of `--stdin`.  I think that fits well with Elm's philosophy of preferring well-documented interfaces.

Also, there is currently behavior where if multiple files are specified, then it is an error to give `--output`.  We should handle this similarly: if `--stdin` is given, then it should be an error to provide any other filenames (and show an appropriate message).

Are there any other edge-cases we need to consider?

eeue56:
Okay, done and done - this makes the code a lot cleaner, so I'm happy with this solution. As it stands, I don't think there's any other edge cases. There might be when more support for tooling is worked on, because of how the tooling wants to work, but I can't think of any right now.

avh4:
Looks great, thank you!

PR #83, opened 2015-12-14
improve readme docs on usage

eeue56:
Updated the README docs with more examples of usage and a header explaining what each one is expected to do


avh4:
I want to make sure the README doesn't get too cluttered, so I will probably simplify this in the future.  But merging it for now.  Thanks!

PR #86, opened 2015-12-18
Update README.md

deadfoxygrandpa:
My plugin has no dependency on the Save On Build plugin to work on every save. It's a setting for Elm Language Support.


avh4:
Thanks! I must have misread your email.

PR #101, opened 2016-01-06
Fix instructions for building from source.

fredcy:
To build from source I had to run `cabal install`. I've updated the README to add that step.


avh4:
Thanks!

Issue #107, opened 2016-01-12
Unicode problems


jvoigtlaender:
The Windows executable (maybe others as well?) suffers from problems we have seen with other Elm tools as well. In a standard Windows shell on my system:
```
> elm-format Main.elm
This will overwrite the following files to use Elmelm-format: <stdout>: commitBuffer: invalid argument (invalid character)
```
A work around is to set the codepage manually:
```
> chcp 65001
```
So one option would be to add this information somewhere in the README or so.

But I think it would be preferable to consider whether Unicode symbols are *really* needed in the console output of this tool.

The specific error above is caused by the apostrophe in [this line](https://github.com/avh4/elm-format/blob/9d8c8d1c5b89810521c53b476d99c8332b68779a/src/Messages/Strings.hs#L17). Is having a Unicode apostrophe instead of a standard `'` there really worth the trouble of telling users about setting codepages? (Not to speak of the fact that not all users will see such a notice, so there will be complaints, unnecessary bug reports, ...)

In fact, on my Windows system, the apostrophe doesn't come out nicely even after changing the codepage:
```
> elm-format Main.elm
This will overwrite the following files to use Elmâ€™s preferred style: 
...
```
So while I don't have an error anymore then, it still looks worse, not better, than using a standard, non-Unicode apostrophe.

avh4:
Fixed in #108.
Agreed, console output should be plain ASCII.

PR #108, opened 2016-01-12
Use standard apostrophe

jvoigtlaender:
Fixes the specific Unicode problem mentioned in https://github.com/avh4/elm-format/issues/107. There might be other instances?

Note also that the file changed here did already contain a standard apostrophe in another console string constant. So there was some inconsistency there anyway.


avh4:
The smart apostrophe was in there by accident.  Thanks for the fix!

PR #111, opened 2016-01-14
Update readme to reflect that atom-beautify 0.28.20 has been released

garetht:


avh4:
That's great news.  Thanks for the PR!

I think the following step "Then in the Atom menu, View -> Reload." can also be removed now.

PR #121, opened 2016-01-24
Three or more double quotes within a multiline string should be escaped

ento:
## Given a test case

```elm
hello = """
""
\"\"\"
\"\"\"\"
"""
```

### before this fix:

```elm
hello = """
""
"""
""""
"""
```

### after this fix:

```elm
hello = """
""
\"\"\"
\"\"\"\"
"""
```


avh4:
This looks good.  The CI failure is non-deterministic and due to a recently-added QuickCheck test that I have a fix for that should be merged soon.

Is there any simpler way to do this without pulling in a new library?

Also, please comment here verifying that you own the copyright to your contributions to elm-format, and please add another commit adding yourself to the `LICENSE` file.

Thanks!

ento:
> Is there any simpler way to do this without pulling in a new library?

Tried using `Data.List.Split`: #122

I hereby verify that I own the copyright to my contributions to elm-format. I've added myself to the license file on both branches, so please take the one that looks better.

// Adding a [CONTRIBUTING](https://github.com/blog/1184-contributing-guidelines) file to the repo might save you some keystrokes in the future!

avh4:
The code for the regex version seems easier to understand.  Do you have any idea which is more efficient?

Janiczek:
Is this how the code should behave? (Eg. how does it find which `"""` ends the multiline string and which is its content?)

If I recall other tools (editors etc.) correctly, they would take the ending `"""` ASAP.

avh4:
@Janiczek the source input `"""\"\"\""""` current formats to `"""""""""`, so this is simply about escaping the internal representation which has already removed the escaping of the source input.

avh4:
I merged this one because the code seemed easier to understand.

Thanks!

ento:
I took some benchmarks, and splitting is consistently more performant than regex, by a factor of 2-10. (but not sure if it matters within the whole scheme of things, i.e. how much time it's going to take up within a single elm-format run)

* ['x' * n + triple quotes](https://cdn.rawgit.com/ento/6b544b06f5032e55eca8/raw/6c1b049065cc6e2a903a8d550b3419710ae06f10/escape-qatend.html) (max n: 10000, order of magnitude: 10ms)
* ['"' * n](https://cdn.rawgit.com/ento/6b544b06f5032e55eca8/raw/6c1b049065cc6e2a903a8d550b3419710ae06f10/escape-qlen.html) (max n: 10, order of magnitude: 10 microseconds)
* ['x' * 10000 with randomly inserted triple quotes * n](https://cdn.rawgit.com/ento/6b544b06f5032e55eca8/raw/6c1b049065cc6e2a903a8d550b3419710ae06f10/escape-qcount.html) (max n: 1000, order of magnitude: 10ms)

benchmark code: https://gist.github.com/ento/6b544b06f5032e55eca8

ento:
All in all: agreed with putting readability first. We can always optimize later if needed. Thanks for the merge!

avh4:
Interesting, thanks for the follow up.  I'll open a new issue with your notes for someone to look at in the future.

PR #127, opened 2016-01-30
update readme to include Homebrew installation

enilsen16:
Merge if & when Homebrew/homebrew#48624 is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,12 @@ elm-format --help  # See other command line options
 
 `elm-format` is still in alpha.  You will need to download the version appropriate for your OS, unzip it, and place `elm-format` or `elm-format.exe` (windows) on your `PATH`.
 
- - Mac: [download](https://github.com/avh4/elm-format/releases/download/0.2.0-alpha/elm-format-0.2.0-alpha-mac-x64.tgz)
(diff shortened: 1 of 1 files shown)


avh4:
Thanks for starting the homebrew PR.  FYI, the release plan is: one more alpha release (0.3) to get feedback from elm-dev, then a beta release (0.4) to get feedback from elm-discuss, and then the stable 1.0 release.

enilsen16:
Homebrew/homebrew-devel-only#41 was merged.

rundis:
:clap:

avh4:
Great, thanks!  

Can you advise on what will need to be done to keep homebrew updated when a new elm-format release is made?  (I expect the 0.3 release to be ready in 2-5 weeks.)

enilsen16:
I have no problem updating it as releases are made...
But basically just fork `homebrew-devel-only` and then change the `link`, `version`, and `sha256` in the devel block of `elm-format.rb`. Homebrew has some more info [here](https://github.com/Homebrew/homebrew/blob/master/CONTRIBUTING.md)

When elm format is finally stable some reorganizing of the formula will need to happen.

PR #144, opened 2016-03-14
Add the ability to run dry runs

eeue56:
As discussed on slack:

- Runs elm format
- Does not modify files
- Prints the names of the files that would change

This will cover what we need for OPs tooling for running efm on our CI!

### DIFF
--- a/src/Flags.hs
+++ b/src/Flags.hs
@@ -13,6 +13,7 @@ data Config = Config
     { _input :: [FilePath]
     , _output :: Maybe FilePath
     , _yes :: Bool
+    , _dry :: Bool
     , _stdin :: Bool
     }
 
@@ -44,7 +45,9 @@ flags =
       <$> Opt.many input
       <*> output
       <*> yes
+      <*> dry
       <*> stdin
+
 
 
 -- HELP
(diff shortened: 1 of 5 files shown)


avh4:
It looks like it will always exit 0.  I think it would be better to call this `--validate` or `--check` or something like that and have it exit non-zero if there are files that would change.

eeue56:
A little harder to do cleanly with the current structure, since it would involve changing where the text is rendered in order to figure out which files have changed. Not impossible, but I think this is the shortest path to achieve the goals that we need (text output saying that something will change, since we can always pipe that into a grep that raises a failure)

avh4:
Can't we render the text as it currently does, but just remember at the end whether to exit 1 or 0?

eeue56:
Oh, hm. Good point, I was thinking of a more complicated solution! I'll give that a go

avh4:
@eeue56 The CLI code could stand some refactoring, but I got the exit code to report the validation status: https://github.com/avh4/elm-format/commit/94f37da84c1310f03dcfa1059ce870b73c94a825

eeue56:
@avh4 I took a similar approach but with ints for encoding more info out about how many files failed

mgold:
@eeue56 Just be sure to clamp that to 127. Some shells use a signed byte for exit codes and you don't want it wrapping. Granted 128 elm files is huge.

eeue56:
It wasn't going to be an exist code, it was going to print the number of files that voliated, then return the exit code if > 0

mgold:
Ok cool.

Issue #139, opened 2016-03-02
Insert newlines for excessively long imports and type annotation


danyx23:
I have a few places where either the list of imported types/functions or the type annotations of a function get very long (in my most extreme case 300+ characters). When I format these things by hand, I insert a newline at every comma in the list of imports or at the arrows of the type annotation, like so:
```elm
import Tasks exposing ( loadPhotoData
                      , loadAudioData
                      , getSlideshowFromServer
                      , downloadAudioFromServer
                      , saveSlideshowToServer
                      , uploadAssetFile
                      , notifyServerOfSuccessfulUpload
                      , resultToString
                      , resultToStringString
                      , slideshowErrorToString
                      , usernameErrorToString
                      , dummyErrorToString
                      , createSlideshowAtServer
                      , registerUsername
                      , createDownloadAudioDataEffect)

notifyServerOfSuccessfulUpload : SlideshowMetadata 
                                 -> AuthToken 
                                 -> Json.Decode.Decoder (Result (RichError c) d) 
                                 -> (Asset a b -> Result (RichError c) d -> EditorAction) 
                                 -> Maybe (Asset a b) 
                                 -> Effects EditorAction
```

Right now, elm-format removes the newlines and creates one long line in both of these cases.

Would it be possible to change elm-format to either 

1. keep the newlines in this case or 
2. introduce (as an option?) the possibility to insert newlines in these two cases if the line length exceeds a threshold

I don't really care about variations of this style (e.g. to keep the arrows with the previous line instead of starting the next line with them), but either of the two options would make elm-format significantly nicer for me :).

avh4:
What I personally think is ideal would be to always add the line breaks, and format it like this:

```elm
import Tasks exposing
  ( loadPhotoData
  , loadAudioData
  , getSlideshowFromServer
  , downloadAudioFromServer
  , saveSlideshowToServer
  , uploadAssetFile
  , notifyServerOfSuccessfulUpload
  , resultToString
  , resultToStringString
  , slideshowErrorToString
  , usernameErrorToString
  , dummyErrorToString
  , createSlideshowAtServer
  , registerUsername
  , createDownloadAudioDataEffect
  )
```

Would people object to that?

avh4:
The rule I'd like to implement here is that if the `exposing` clause contains any newlines, then it will put each list item on a separate line, otherwise keep it all on one line.  This would be consistent with most other formatting in elm-format.

Also, I think the style choice that makes the most sense is

```elm
[...]

Issue #142, opened 2016-03-08
Keep newlines for long type annotations


avh4:
Split from #139 

```elm
notifyServerOfSuccessfulUpload : SlideshowMetadata 
                                 -> AuthToken 
                                 -> Json.Decode.Decoder (Result (RichError c) d) 
                                 -> (Asset a b -> Result (RichError c) d -> EditorAction) 
                                 -> Maybe (Asset a b) 
                                 -> Effects EditorAction
```

avh4:
The formatting should match other multiline type annotations, e.g.

```elm
notifyServerOfSuccessfulUpload :
  SlideshowMetadata
  -> AuthToken 
  -> Json.Decode.Decoder (Result (RichError c) d) 
  -> (Asset a b -> Result (RichError c) d -> EditorAction) 
  -> Maybe (Asset a b) 
  -> Effects EditorAction
```

avh4:
Fixed in 9079a11.

PR #147, opened 2016-03-17
Import multiline #139

danyx23:
This is my first stab at #139, getting multiline "import .. exposing()" lists to be formatted as multiline by elm-format.

This is my first PR in Haskell, so please tell me if you don't like anything or I should change something. 

I added a few test cases as you suggested and managed to get elm-format to create the according output.

There are two things I wasn't entirely sure about:

1. in parser/src/Parse/Module.hs, around line 112, I use pushNewlineContext before the choice parser. I think that this will wrongly identify import constructs as multiline if a line break occurs as part of the ".." option, but I wasn't sure what the best other way to deal with this would be. Please let me know if you think this is a problem.

2. At the output generation, I wasn't sure how best to deal with a non-SingleLine version of name, so I duplicated the 2 relevant cases and tried my best to accommodate them.

Thank you for your work!

### DIFF
--- a/parser/src/AST/Module.hs
+++ b/parser/src/AST/Module.hs
@@ -60,6 +60,6 @@ data UserImport
 
 data ImportMethod = ImportMethod
     { alias :: Maybe (Comments, PreCommented String)
-    , exposedVars :: (Comments, PreCommented (Var.Listing Var.Value))
+    , exposedVars :: (Comments, PreCommented (Var.Listing Var.Value), Bool)
     }
     deriving (Eq, Show)
--- a/parser/src/Parse/Module.hs
+++ b/parser/src/Parse/Module.hs
@@ -109,19 +109,16 @@ listing item =
   expecting "a listing of values and types to expose, like (..)" $
   do  (_, preParen) <- try (whitespace <* char '(')
       (_, pre) <- whitespace
+      pushNewlineContext
       listing <-
           choice
-            [ (\_ pre post -> Var.OpenListing (Commented pre () post)) <$> string ".."
-            , (\x pre post -> 
-               {- do
-                  pushNewlineContext
-                  sawNewline <- popNewlineContext
-                return $ -}
-                      Var.ExplicitListing (x pre post)) <$> commaSep1' item
+            [ (\_ pre post -> (Var.OpenListing (Commented pre () post))) <$> string ".."
+            , (\x pre post -> (Var.ExplicitListing (x pre post))) <$> commaSep1' item
             ]
+      sawNewline <- popNewlineContext
       (_, post) <- whitespace
       char ')'
-      return $ (preParen, listing pre post, False)
+      return $ (preParen, listing pre post, sawNewline)
 
 
 value :: IParser Var.Value
--- a/parser/src/Parse/Module.hs
+++ b/parser/src/Parse/Module.hs
@@ -61,7 +61,7 @@ moduleDecl =
       (_, preName) <- whitespace
       names <- dotSep1 capVar <?> "the name of this module"
       (_, postName1) <- whitespace
-      (postName2, exports) <- option ([], Var.OpenListing (Commented [] () [])) (listing value)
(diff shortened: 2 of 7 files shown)


src/ElmFormat/Render/Box.hs
```diff
@@ -394,15 +417,15 @@ formatImport aimport =
                           ( name', Just as', Just exposing' ) ->
                             stack1
                               [ line $ keyword "import"
-                              , indent name'
-                              , indent $ indent as'
-                              , indent $ indent exposing'
+                              , indent $ name'
+                              , indent $ indent $ as'
+                              , indent $ indent $ exposing'
```
mgold: These new `$`s aren't necessary, as is the one on 428. Haskell `$` is Elm's `<|`.

parser/src/AST/Module.hs
```diff
@@ -60,6 +60,6 @@ data UserImport
 
 data ImportMethod = ImportMethod
     { alias :: Maybe (Comments, PreCommented String)
-    , exposedVars :: (Comments, PreCommented (Var.Listing Var.Value))
+    , exposedVars :: (Comments, PreCommented (Var.Listing Var.Value), Bool)
```
mgold: This is hardly unique to this PR but I wish there was a way to document that the bool signifies whether the import is multiline or not.

src/ElmFormat/Render/Box.hs
danyx23: Ah of course! Thanks for pointing this out.

parser/src/AST/Module.hs
danyx23: Yes I was thinking if using newtype to make this more clear and avoid accidents might be a good idea but I didn't want to mix that into the work on this PR. If there is agreement on this I could create an issue for this and convert the projects multi line Bools to the new type

parser/src/Parse/Module.hs
```diff
@@ -96,27 +96,29 @@ import' =
...
+          return (preExposing, (postExposing ++ postExposing2, listing'), multiline)
 
 
-listing :: IParser a -> IParser (Comments, Var.Listing a)
+listing :: IParser a -> IParser (Comments, Var.Listing a, Bool)
 listing item =
   expecting "a listing of values and types to expose, like (..)" $
   do  (_, preParen) <- try (whitespace <* char '(')
       (_, pre) <- whitespace
+      pushNewlineContext
```
avh4: To be consistent with the rest of elm-format, I think pushNewlineContext should happen a line earlier (before the whitespace).  I will make that change and add a commit.

parser/src/Parse/Module.hs
```diff
@@ -96,27 +96,29 @@ import' =
...
       listing <-
           choice
-            [ (\_ pre post -> Var.OpenListing (Commented pre () post)) <$> string ".."
-            , (\x pre post -> Var.ExplicitListing (x pre post)) <$> commaSep1' item
+            [ (\_ pre post -> (Var.OpenListing (Commented pre () post))) <$> string ".."
+            , (\x pre post -> (Var.ExplicitListing (x pre post))) <$> commaSep1' item
             ]
+      sawNewline <- popNewlineContext
       (_, post) <- whitespace
```
avh4: Same here, popNewlineContext should move down one line.  I will make this change as well.

avh4:
Looks great, thanks.  I will add a commit for the two comments I made.

avh4:
@danyx23 Since this is your first contribution, can you please do the following first:

Add another commit adding yourself to /LICENSE
Confirm via a comment on this PR that you are the owner of the code you are contributing to elm-format, and that you are licensing it under the BSD3 license given in /LICENSE

Thanks!

danyx23:
I confirm that I am the owner of the code I am contributing to elm-format. All code contributed by me to elm-format is licensed under the BSD3 license given in /LICENSE.

I have added myself above Marica to avoid a trivial merge conflict with the master branch where Max added himself in the same line :)

I will try to tackle #142 in a similar manner to this one in a separate PR

avh4:
Sorry for the delay in getting back to this.  Merged!

PR #160, opened 2016-04-26
change 'necessary' to 'necessarily'

nicklawls:
in context: "It makes code **easier to maintain**, because you can no longer have diffs related only to formatting; every diff necessarily involves a material change"


avh4:
Thanks!

PR #216, opened 2016-07-15
Update install section for Emacs

marcowahl:
Added description on how to configure Emacs to elm-format on every save.


avh4:
Thanks!

PR #217, opened 2016-07-26
Add WebStorm setup instructions to Readme

colinf:


avh4:
Thanks!   This should also work for all JetBrains IDEs (like RubyMine, PyCharm, IntelliJ), right?

Janiczek:
It should.

I wonder if it can be made more satisfactory (these options make the cursor jump around quite a lot and don't let you type stuff unless you're very fast).

I've settled on defining an `External Tool` elm-format and having a keyboard shortcut for running that. This way I can emulate the "format on save" behaviour from other editors where I write the program unharmed and press a keyboard shortcut to format.

external-tool

keymap

colinf:
Hi Martin,

I've only tried it in WebStorm but yes I agree that as far as I know it should work fine in the other JetBrains IDEs as well.

The settings I posted are working fine for me on my 4 year old Mac and they format on save without any additional key commands etc. I'm not getting delays or jumping cursors at all so I'm not really sure what the problem us, but one thing to check is that you don't have the "Immediate file synchronisation" setting checked as this could make the UI unresponsive as it tries to reformat on every keystroke rather than only formatting on save.

Cheers,
  Colin

PR #221, opened 2016-08-16
Output validation result in a machine-readable format

ento:
## Motivation

We run `elm-format --validate` in our CI builds to enforce standard code formatting. Being able to save the validation result in a standard machine-friendly format allows a tighter integration with the CI server (Jenkins in our case), leading to better visibility of which files failed the build for what reason.
## Intended usage

```
elm-format --validate --format json app spec
```

To convert to checkstyle's xml format:

```
elm-format --validate --format json app spec | jq -n -r -f checkstyle.jq > elm-format-validation-result.xml
```

checkstyle.jq is from [this gist](https://gist.github.com/ento/a4e2a251f54fe531ede9a8cd1960c7d6#file-checkstyle-jq). I'll push it to [elm-ops-tooling](https://github.com/NoRedInk/elm-ops-tooling) once this PR is merged.
## Limitations, concerns and notes
- Only reports which files failed validation, not the exact lines that failed
- Having a `--format` option for a binary called `x-format` might be confusing
- Code organization inspired by [shellcheck](https://github.com/koalaman/shellcheck/tree/dbafbb3b3b0e6d8d04b99f42f8816dada44fe216/ShellCheck/Formatter)
## Input/Output sample

``` sh
$ cp "tests/test-files/transform/Examples.elm" "_input.elm"
$ cp "tests/test-files/transform/Examples.elm" "_input2.elm"
$ ./dist/build/elm-format-0.17/elm-format-0.17 _input.elm _input2.elm --validate --format json
```

outputs:

``` json
{"path":"_input2.elm","line":1,"message":"File would be changed","column":1}
{"path":"_input.elm","line":1,"message":"File would be changed","column":1}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /formatted.elm
 /_input.elm
 /_input2.elm
+/_stdout.txt
 /.cabal-sandbox/
 *.tix
 /.stack-work/
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ before_install:
 install:
 - stack setup
 - stack build --only-dependencies
+- npm install -g ajv-cli
 
 
 script:
--- a/README.md
+++ b/README.md
@@ -308,5 +308,6 @@ stack build
 
 ```bash
 brew install shellcheck
+node install -g ajv-cli
 ./tests/run-tests.sh
 ```
--- a/src/CommandLine/Helpers.hs
+++ b/src/CommandLine/Helpers.hs
@@ -3,7 +3,7 @@ module CommandLine.Helpers where
 import System.IO
 import System.Exit (exitFailure, exitSuccess)
 import Messages.Types (ErrorMessage(..))
-import Messages.Strings (renderErrorMessage)
+import Messages.Strings (showErrorMessage)
 
 import qualified Reporting.Annotation as RA
 import qualified Reporting.Report as Report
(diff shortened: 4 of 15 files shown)


avh4:
I think it's good to have an output option that gives machine-readable info about what failed.  But I'm hesitating on this because I'm uncertain about having checkstyle XML be that format.

What are your thoughts about having elm-format produce a JSON report, and then there could be an external script that could transform the json into XML?

ento:
JSON as our choice of machine-readable format sounds good. jq appears to be versatile enough to transform a JSON array to either a file-by-file format (checkstyle-style) or a line-by-line format (ala pep8, for example): https://gist.github.com/ento/a4e2a251f54fe531ede9a8cd1960c7d6

I'll make the necessary changes to do the JSON instead.

```
[
  { "path": "path/to/Main.elm",
    "line": 1,
    "column": 1,
    "message": "<file would change>"},
  { "path": "path/to/View.elm",
    "line": 1,
    "column": 1,
    "message": "file would change"}
]
```

^let me know of any concerns with this JSON structure.

ento:
Changed the format to one-object-per-line, because commas are hard. (need to make the formatter a stateful thing, or pass in the index of the file being processed to avoid an invalid trailing comma at the end)

```json
{ "path": "path/to/Main'.elm", "line": 1, "column": 1, "message": "<file would change>"}
{ "path": "path/to/Main'.elm", "line": 2, "column": 1, "message": "file would change"}
{ "path": "path/to/View.elm", "line": 1, "column": 1, "message": "file would change"}
```

ento:
@avh4 ready for another look

avh4:
Rebased as https://github.com/avh4/elm-format/pull/316

PR #225, opened 2016-08-26
atom-elm-format guideline changes

benjick:
Since 2.0 the default is to format on save :tada:

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Find your editor in the table below.  The recommended plugin for each editor is 
     <td rowspan=2><a href="https://atom.io/">Atom</a></td>
     <td>:trophy: <a href="https://atom.io/packages/elm-format">atom-elm-format</a></td>
     <td>:white_check_mark: <a href="#atom-elm-format-installation">2 steps</a></td>
-    <td>:warning: must use format-on-save or save file before formatting</td>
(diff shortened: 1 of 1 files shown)


avh4:
Great, thanks for the update!

avh4:
@benjick I actually just reverted this.... having format-on-save be the default is the existing checkmark in the "Format on Save" column.

The warning in the "Formatting" column is because is because of this:

> When formatting a file, if the in-memory version of the file and the on-disk version of the file differ, then the plugin should format the in-memory version of the file. (Saving the file before formatting is also acceptable.)

The behavior when I tested it was that if format-on-save is turned off and the file is not saved before running the manual formatting command, then there are some problems.

benjick:
Oh okay, now I understand. Sorry about that. I will look into making it
compatible.

Which is preferred? Saving the file or just replacing the buffer?

benjick:
Okay I read the guidelines now, will look into this!

avh4:
@benjick I assumed different things might be conventional for different editors, so didn't want to dictate that.  Maybe check out popular atom formatting plugins for other languages and see what's common?

benjick:
The thing I do now I borrowed from eslint, but I think it's more important
to be compliant with the plug-in itself, than other plug-ins for the editor

PR #228, opened 2016-09-02
simplify vs code editor section

devinrhode2:
The `elm` tools package provides more functionality than the elm-format plugin, so I removed this duplicate option. I think it's safe to say most people would prefer or at least be ok with the extra options this package provides.


avh4:
Thanks for the PR.  I want to understand the plugins a bit more before merging this.

I think it's good to include in the list all existing plugins that someone might come across, so that if they happen to hear about another plugin from some other source, they can see in the elm-format readme how it compares.  However, if the plugin is clearly deprecated or no longer available, then I think it makes sense to remove it completely.

avh4:
I think this got resolved with another PR that got merged.  Thanks for this!

PR #235, opened 2016-09-22
Clarified formatOnSave instructions for VSCode

jmandzik:


avh4:
Thanks for the improvement!

Issue #84, opened 2015-12-14
elm-format distribution 


eeue56:
In the long term, it makes sense to me for `elm-format` to be included with Elm platform. This is what gofmt did, after all. However, I can also see some benefit to having it being a separate package, installable via npm. I would also like to make a node interface to it, so that I can hook it up to webpack and enforce elm-formatted code regardless of platform.

I'm happy to do the leg work here. Just wanted to hear your thoughts @avh4

avh4:
Yes, I'm hoping to include it in Elm platform 0.17 or the version after.

I'm not a fan of the npm wrappers/installers for elm-test and elm-compiler.  (I'm more interested in getting proper homebrew/apt-get/chocolatey packages created for elm-platform.)  But if you make one, I'd be happy to link to it in the README.  If there are any changes to the CLI that would make it easier to use in webpack etc plugins, suggest them in separate issues.

eeue56:
Okay, yeah, I'm going to go ahead with this - the benefit of having an npm wrapper is that we can get it to install binaries local to a project when we pack them with webpack etc.

I'll let you know when I have something that works, we're probably gonna have a stab at this in the next couple of weeks or so.

rtfeldman:
I think in an ideal world you have `npm install -g elm` and no other wrappers needed.

The other nice thing about an `npm` wrapper for Elm itself is that it means you can one-line install Elm onto any CI server running Node (which you already need in order to execute `elm-test`).

But yeah, having both `elm-test` and `elm-format` ship with the Elm platform would be :metal:

eeue56:
That is actually something else Golang does very well - it ships with a `go test` command, which will look for files with the name `Test<x>.go` in the current project and run them with the tester. It's an awesome way of encouraging TDD - it's a lot easier to get started with tests if you don't have to first download all sorts and set up a suite

rtfeldman:
@eeue56 I like that idea! We can't quite follow the same algorithm because we need a separate `elm-package.json`, but being able to run `elm test` on a fresh project and have it Just Work seems like a noble goal. Mind opening an issue on https://github.com/rtfeldman/node-elm-test for that?

eeue56:
Done! https://github.com/rtfeldman/node-elm-test/issues/21

lpil:
Given Elm, Elm Test, and Elm Oracle seem to have settle on npm for now, would it make sense to also package Elm Format on npm?

avh4:
[...]

PR #237, opened 2016-09-25
Add npm installer

bobbypriambodo:
As stated in #84, here is the result of my fiddling with npm installer for `elm-format`. I'm drawing many examples from [`elm-platform` npm installers](https://github.com/elm-lang/elm-platform/tree/master/installers/npm). I'm putting the installer at `package/npm` directory (may be moved if you wanted to).

Some questions, though:
1. I was thinking of creating my own repo rather than forking and creating a dir in this repo's `package`, but decided not to just to keep it centralized. Should I have done that instead?
2. I still have no idea what to export for local npm installation and use by `require`. Will exporting the binary paths (like `elm-platform` does) be enough?
3. I've referred the publishing process to the npm official docs on the README (along with some notes on version bumping). Please tell me if I need to add something more.

I've tested this by running `npm install -g .` inside the `package/npm` directory and it successfully downloaded and installed `elm-format` to my global module. Not publishing it to npm yet since npm have a kind of [restrictive publishing policy](http://blog.npmjs.org/post/141905368000/changes-to-npms-unpublish-policy).

Kindly take a look? @avh4 @eeue56

### DIFF
--- a/package/npm/.eslintrc.yml
+++ b/package/npm/.eslintrc.yml
@@ -1,0 +1,6 @@
+extends: eslint:recommended
+env:
+  node: true
+  es6: true
+rules:
+  no-console: off
--- a/package/npm/.gitignore
+++ b/package/npm/.gitignore
@@ -1,0 +1,2 @@
+node_modules/
+bin/
--- a/package/npm/.gitignore
+++ b/package/npm/.gitignore
@@ -1,2 +1,3 @@
 node_modules/
 bin/
+*.log
--- a/package/npm/README.md
+++ b/package/npm/README.md
@@ -1,0 +1,20 @@
+npm package for elm-format
+==========================
+
+This package installs [`elm-format`](https://github.com/avh4/elm-format) via [`npm`](https://www.npmjs.com/).
+
+## Usage
+
+```
+npm install -g elm-format@0.4.0-alpha
+```
+
+This will install the binaries `elm-format` and `elm-format-0.16` to your global `node_modules`.
+
+## Publishing
+
+Publishing npm packages follows the [procedure stated in the official npm docs](https://docs.npmjs.com/getting-started/publishing-npm-packages).
+
+### Bumping `elm-format` version
+
+The downloaded binary version will follow the one stated in `package.json`. If `elm format` is updated, this package can be bumped by only updating the `version` field in the `package.json` (given that the download url structure does not change).
--- a/package/npm/binwrap.js
+++ b/package/npm/binwrap.js
@@ -8,9 +8,11 @@ module.exports = function(executable) {
 
   const os = process.platform;
 
-  const extension = os === 'win32' ? '.exe' : '';
-  const filename = `${executable}${extension}`;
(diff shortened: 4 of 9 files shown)


bobbypriambodo:
Sorry, this is still broken in Windows (`zlib` and `unzip` behaves differently). I'm still working on it.

bobbypriambodo:
I think I've fixed the Windows problem, at least now it will point to the correct `.exe` file. Unfortunately I have no Windows machine to test this at the moment, can anyone check if it works?

bobbypriambodo:
I can confirm that it works on Windows 10 using PowerShell (see screenshot below), while CMD runs `elm-format` just fine but is treating the `.16` in `elm-format-0.16` as an extension... Any idea on the expected behavior on Windows?

screen shot 2016-09-25 at 5 35 54 pm

hastebrot:
This works on Ubuntu 16.04 (x64). Thanks.

~~~
$ hub checkout https://github.com/avh4/elm-format/pull/237
$ cd package/npm/
$ sudo npm install -g
$ which elm-format
/usr/bin/elm-format
~~~

(using `hub` to checkout your PR: https://hub.github.com/)

felixbecker:
Hi, don't know if this is the right place to post my feedback on an issue.  
I got an issue with the npm elm-format installer on windows 7 with the gits MINGW64 bash. 

It seems that Windows or MINGW64 has a problem with executables without an extension. Since elm-format is a cmd file it doesn't work by executing `elm-format` but does successfully if you execute `elm-format.cmd`. But not really sure here.

if I execute `elm-format` without the extension (.cmd) I get the following message 
`/d/Users/TheUserIAm/AppData/Roaming/npm/elm-format: line 1: /node_modules/elm-format/binaries/elm-format: No such file or directory`. 
Only on gits MINGW64 bash on windows if I use the regular CMD it works.
Any thoughts here? (using elm-format for elm 0.18)

bobbypriambodo:
@felixbecker Where did you get elm-format installer from? If it's from npm (i.e. via `npm i elm-format`), then I believe it should be @rtfeldman's implementation from #288. That PR is more recent and I think is the one that has @avh4's blessing. I'll wait a bit, but if there's no further instruction I'll close this PR.

felixbecker:
Yes, I ran `npm install elm-format -g` should I post this topic somewhere else?

avh4:
@felixbecker yes, can you continue this discussion on #288 

For your case, I would expect the binaries to be `elm-format.exe`, not `elm-format.cmd`.

rtfeldman:
@felixbecker I suspect the problem is that the installer asks `node` what the current platform and architecture are, and fetches the appropriate binary. You're getting the Linux version instead of the Windows `.exe` file because it's self-reporting as Linux from `bash`.

Does that sound plausible?

avh4:
Thanks for the detailed work on this!  I mostly went with the approach in https://github.com/avh4/elm-format/pull/288, but a took a couple small things from here.

PR #244, opened 2016-09-29
Make it possible for (haskell)stack to handle project

kuznero:


src/Box.hs
```diff
@@ -247,7 +247,7 @@ initRow startColumn =
 
 spacesInTab :: Int
 spacesInTab =
-  4
+  2
```
kuznero: Correct, sorry that was done after I created this PR. Should have branched out.

supermario: It's funny that this is here actually – I came here with the following train of thought:

- We'd like to use 2 spaces instead of 4, can I configure it?
- Hrm, can't configure... maybe I can do my own `elm-format` build
- How do I build this using stack then...
- Oh, a PR about stack
- Heh! 2 spaces 😁

kuznero: I think number of spaces per tab should be configured. 4 is ok as a default value, but it would be fair to let others to choose...

avh4:
Thanks!  I think this is fixed in https://github.com/avh4/elm-format/commit/a91dc8374f864140bc329f9fb47ed79c6e1edd33 (I kept the strict version numbers for dependencies, and also fixed up CI and the README to use stack.)

kuznero:
Thanks!

PR #245, opened 2016-09-30
Details for adding user keymap entries in lighttable

pythe:
I'm new to Lighttable, so hopefully these improved instructions would work for other novices.

After spending a few hours improving my Lighttable configuration, I found that using the on-save hook was more correct than re-binding ctrl-s. This allows users who have save-on-focus-lost configured to benefit from formatting as well.


pythe:
Okay, this isn't ready yet. I'll close it and reopen when I've actually got it working.

PR #248, opened 2016-10-10
Add File Watchers plugin installation step to JetBrains IDE installation

francisdb:


avh4:
Thanks for the update!

PR #252, opened 2016-10-18
Enhancement to LightTable installation instructions.

nukecoder:
Enhancement to LightTable installation instructions to make it more explicit for new users.


avh4:
Thanks for the update!

PR #261, opened 2016-11-02
Make it more clear that multiple Elm files can be formatted as a batch

klazuka:
The CLI help docs made it seem like elm-format could only be invoked
against a single Elm source file, but the README includes an example
of how elm-format can be used to recursively format all .elm files
in a directory. This change brings the CLI help docs in-line with the
README and the capabilities of the tool.

### DIFF
--- a/src/Flags.hs
+++ b/src/Flags.hs
@@ -95,7 +95,7 @@ helpInfo defaultVersion =
     mconcat
         [ Opt.fullDesc
         , Opt.header top
-        , Opt.progDesc "Format an Elm source file."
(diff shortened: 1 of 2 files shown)


avh4:
Looks good, but the file `tests/ElmFormat/CliTest/Usage.stdout` needs to be updated to match.  Thanks!

klazuka:
Doh. Didn't even think that those strings would be under test. Sorry about that.

avh4:
Great, thanks!  This will be in the next release

PR #279, opened 2016-11-22
Sorted imports and exposed variables

basile-henry:
This is just a suggestion for elm-format, not fixing any of the issues (as far as I'm aware).

This pull request implements a sorting of the imports a bit like [stylish-haskell](https://github.com/jaspervdj/stylish-haskell) does for Haskell. I think this goes well with the spirit of elm-format to have a unique way to format elm source code in this case making it easier for a programmer to find the relevant imports quickly. This is especially useful when the list of imports is big.

The sorting is done on the alias given to the module and when no alias have been declared it uses the module name.

When the imports are separated by a comment it will keep them separated in the same way for example:

```elm
import String
import Json.Decode as Json
import List exposing (..)
-- Local imports
import Utils
import MyAwesomeModule
```

becomes:

```elm
import Json.Decode as Json
import List exposing (..)
import String


-- Local imports

import MyAwesomeModule
import Utils
```

This should work on all the versions of Elm supported by elm-format (but I haven't test this extensively).

What do you think of this? Would it be useful?

### DIFF
--- a/elm-format.cabal
+++ b/elm-format.cabal
@@ -53,6 +53,7 @@ library
         Elm.Utils
         ElmFormat.Cli
         ElmFormat.Parse
+        ElmFormat.Render.Box
         ElmFormat.Render.ElmStructure
         ElmFormat.Render.Text
         ElmFormat.World
@@ -68,7 +69,6 @@ library
     other-modules:
         CommandLine.Helpers
         ElmFormat.Filesystem
-        ElmFormat.Render.Box
         Flags
         Messages.Strings
         Messages.Types
--- a/parser/src/AST/Variable.hs
+++ b/parser/src/AST/Variable.hs
@@ -26,3 +26,12 @@ data Value
     | OpValue SymbolIdentifier
     | Union (PostCommented UppercaseIdentifier) (Listing UppercaseIdentifier)
     deriving (Eq, Show)
+
+instance Ord Value where
+  (Value a)        <= (Value b)        = a <= b
+  (OpValue a)      <= (OpValue b)      = a <= b
+  (Union (a, _) _) <= (Union (b, _) _) = a <= b
+  (Union _ _)      <= _                = True
+  _                <= (Union _ _)      = False
+  (OpValue _)      <= (Value _)        = True
+  (Value _)        <= (OpValue _)      = False
--- a/src/ElmFormat/Render/Box.hs
+++ b/src/ElmFormat/Render/Box.hs
@@ -552,43 +552,30 @@ formatImport elmVersion aimport =
             formatComment c
 
 sortImports :: [AST.Module.UserImport] -> [AST.Module.UserImport]
-sortImports imports =
-    let
-        importIsComment :: AST.Module.UserImport -> Bool
-        importIsComment aimport =
-            case aimport of
-                AST.Module.UserImport _ -> False
-                AST.Module.ImportComment _ -> True
-
-        aliasOrName :: AST.Module.UserImport -> [UppercaseIdentifier]
+sortImports = sortGroupsOn aliasOrName
+    where
+        aliasOrName :: AST.Module.UserImport -> Maybe [UppercaseIdentifier]
         aliasOrName aimport =
             case aimport of
                 AST.Module.UserImport uimport ->
                     case RA.drop uimport of
                         ((_, identifiers), importMethod) ->
                             case AST.Module.alias importMethod of
-                                Just (_, (_, alias)) -> alias : identifiers
-                                Nothing -> identifiers
-                AST.Module.ImportComment _ -> []
+                                Just (_, (_, alias)) ->
+                                    Just $ alias : UppercaseIdentifier "" : identifiers
 
-        onInit :: ([a] -> [a]) -> [a] -> [a]
-        onInit f xs = f (List.init xs) ++ [List.last xs]
+                                Nothing ->
+                                    Just identifiers
 
-        importGroupMap f groups =
-            case groups of
-                [] -> []
-                [x] ->
-                    case last x of
-                        AST.Module.ImportComment _ -> [onInit f x]
-                        AST.Module.UserImport _ -> [f x]
-                (x:xs) -> onInit f x : importGroupMap f xs
+                AST.Module.ImportComment _ -> Nothing
 
-        sortImports' = List.sortBy (compare `Function.on` aliasOrName)
-    in
-        imports
-            |> splitWhere importIsComment
-            |> importGroupMap sortImports'
-            |> concat
+sortGroupsOn :: Ord b => (a -> Maybe b) -> [a] -> [a]
+sortGroupsOn f list =
+    list
+        |> zip (map f list)
+        |> List.groupBy ((==) `Function.on` (Maybe.isJust . fst))
+        |> concatMap (List.sortBy (compare `Function.on` fst))
+        |> map snd
 
 formatListing :: (a -> Box) -> AST.Variable.Listing a -> Maybe Box
 formatListing format listing =
--- a/src/ElmFormat/Render/Box.hs
+++ b/src/ElmFormat/Render/Box.hs
@@ -412,7 +412,9 @@ formatImport elmVersion aimport =
                           formatImportClause
                             (formatListing (formatVarValue elmVersion))
                             "exposing"
-                            (exposingPreKeyword, exposingPostKeywordAndListing)
+                            ( exposingPreKeyword
+                            , fmap sortListing exposingPostKeywordAndListing
+                            )
 
                         formatImportClause :: (a -> Maybe Box) -> String -> (Comments, (Comments, a)) -> Maybe Box
                         formatImportClause format keyw input =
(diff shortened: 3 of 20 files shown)


basile-henry:
I have only just noticed that this has been discussed to great lengths without really coming to a conclusion in #8.

It looks like there are many different views regarding this feature and this pull request does cover some of the requests:

- Sorting alphabetically by group
- Groups have to be separated by a comment (not just a newline) which might not be an issue as @mordrax pointed out:

  > currently, elm-format puts spaces around single line comments --, if these were used to 'group' imports, we'd not be creating another convention?
- Alphabetisation done the easy way which might be good enough as was suggested by @rtfeldman:

  > I believe "sort using Haskell's default Ord implementation for String in EN-US" would fit the bill. 😃

What this pull request implements which hasn't been discussed so far is the use of the alias given to the module name for the sorting. What do you think of this behaviour? Is it counter-intuitive, useful?

Rolograaf:
Why is `Events` above `Html.App`?

basile-henry:
Are you referring to this?

```elm
import Html.Events exposing (..)
import Html.App as Html
```

This is due to the fact that it is also using the alias for the sorting. This is the sort of extra feature I wanted to discuss and I am completely open to removing if it is counter-intuitive.

The advantage of using the alias is so the modules are sorted by the names they're actually used with in the program. When multiple modules share the same alias (which can be useful to only have one name space for multiple modules) then the module name is taken into account for the sorting.

For example (not that the module names make that much sense) :

```elm
import Html
import List
import Weird.Html as Html
import Custom.Html as Html
import Custom.Tree
```

become:

```elm
import Custom.Tree
import Html
import Custom.Html as Html
import Weird.Html as Html
import List
```

In the case you're referring to, this pull-request is comparing ["Html", "Events"] with ["Html", "Html", "App"] (alias and then module name) which results in the sorting shown above.

Would you rather everything was sorted by module name rather than use the alias?

Rolograaf:
All clear! So actually `Html.App` could have alias `Program` ?

Rolograaf:
Will it also sort the exposing functions when more than one?

basile-henry:
Yes `Html.App` could have any alias you like and the alias would be used for the sorting.

This first question you had actually gave me an idea, would you prefer if the alias was always just below the module of the same name?

```elm
import Html exposing (..)
import Html.App as Html
import Html.Lazy as Html
import Html.Attributes exposing (..)
import Html.Events exposing (..)
```

Instead of just prepending the alias the module name, this could be implemented by simply adding a space in between the alias and the module name to give some sort of "priority" in the sorting.

`["Html", "", "Html", "App"]` would then be above `["Html", "Events"]`.

basile-henry:
I haven't thought about the exposed functions, definitely the next step! stylish-haskell already does that I believe. :-)

Rolograaf:
@avh4 does have a strong opinion about the sequence of exposed functions (bravo elmlive), but my brain wants it sorted alphabetically... please note weight of my vote is featherlight!

basile-henry:
If I were to sort the exposed function alphabetically I should probably keep the same grouping behaviour as for the imports. Only sort within a group and separate groups by comments.

Would this go against @avh4 exposed functions sequence preferences? If comments are written, the programmer can still have full control of the order should they want to.

basile-henry:
I've experimented a bit with the sorting of exposed variables which should be fairly straight forward. There is an issue to do with the comments though. Since all that exposed variables can be commented (that is both pre-commented and post-commented, I am unsure which comments should follow which exposed variable when sorting. For inline comments this is not an issue as it depends on which side of the commas the comments are. It is an issue for the single line comments in the multiline exposed variables format (which looks really nice by the way).

With the code I currently have it basically keeps comments as if they were post-comments which makes sense from a programming perspective but probably isn't the behaviour wanted for someone using elm-format.

I will push the commit implementing sorting of the exposed variables here as well so we can talk about specifics more easily but I can change the commit or remove it entirely if such feature is unwanted or we can't agree on the wanted behaviour.

Rolograaf:
Sorry, not using comments a lot and never between exposed functions or types. I'm sure linters, formatters, parsers and compilers should cater for every use-case, but for me personally a good name does not need further comments.

avh4:
I think there are a few separate issues here:

I'd like to have a separate discussion for each part (and then come back to this PR once some of the decisions are made).

 - should import lines be sorted?  That discussion can continue at #8 (including the additional questions of how should grouping work, and should it sort by the "as" name)
 - should exposings in import lines be sorted?  We should make a discussion issue for that
 - should exposings in the module line be sorted?  (This is what I've referred to in some of my videos)

Sorry for being slow on this, but I think this is a change that would impact a lot of people.

basile-henry:
> Sorry for being slow on this, but I think this is a change that would impact a lot of people.

No problem, this should obviously be a community decision and it should be tailored to benefit most people. @avh4 how have you tackled this kind of decision in the past? Simply by voting in the issue threads? Are you looking for specific arguments/use cases for or against such changes?

> I'd like to have a separate discussion for each part

Would it be helpful if I split the commits for the different parts in different branches/pull requests?

> should exposings in the module line be sorted?

This pull request does not currently change the module line at all. This can be implemented as well but I can see that being an issue. How are the elm docs generated? Is it like for a Haskell package where the order in which the docs for each function appear depends on the order they are exposed?

avh4:
> No problem, this should obviously be a community decision and it should be tailored to benefit most people. @avh4 how have you tackled this kind of decision in the past? Simply by voting in the issue threads? Are you looking for specific arguments/use cases for or against such changes?

Most of the decisions thus far in elm-format have been for things where making the change didn't really matter to anyone.  In the case of import sorting, it wasn't done yet because there were clear examples of people who wanted to group/organize their imports in a specific way.  So we should find out if anyone still cares about manually organizing their imports, and if so, can sorting be done in a way that works with those desires, and if not, is sorting a big enough benefit that we want to prevent people from not sorting manually.

Rolograaf:
BTW I have noticed that package install also does alphabetical sorting in the `elm-package.json` file for dependencies. However I do not think anyone is bothered much there in preferences or grouping since a lot of mutations on that settings-file happen automagically in elm.

rtfeldman:
> How are the elm docs generated? Is it like for a Haskell package where the order in which the docs for each function appear depends on the order they are exposed?

No, it's based on the `@docs` annotations in the doc string at the top of the file. So reordering `exposes` would not affect docs.

basile-henry:
I have just added a new commit to do the sorting of the exposed variables in the header and I have rebased this pull request on the latest master.

I am adding this new sorting so we can discuss it more easily, by no means does it have to be accepted with the rest of the pull request. Should only part of this pull request be of interest, I'll happily split it up or create a new pull request with the subset of features.

The sorting of the module exposed variables uses the same sorting mechanism as for the imports exposed variables and depends on these commits. I have also updated the tests to reflect these changes (look at the `Arbitrary` instance for `Module`). I am unsure where the `sortListing` function should go in elm-format modules so I left it where it is used in `ElmFormat.Render.Box` and added the module to the exposed module for testing purposes.

@avh4: On the subject of testing: instead of exposing all the internal modules for testing purposes in elm-format.cabal we could simply add the relevant source directories to the test suite. This has the advantage of giving access to all the modules to the test suite while keeping these modules from being accessed from outside the elm-format package.

avh4:
This is implemented in e4579233c0c99c23dbb0f2d0b8d7ba0fa856affc...9c8d11bb7064e7c36c661d6c8204266e5b81468e.

@basile-henry I didn't end up using your code because I wanted to do this in a way that would also merge duplicate imports, and that was the direction I wanted to move the AST in anyway, but thank you for pushing this forward and doing the initial work.

stil4m:
@avh4 On the note that duplicate imports also get merged. This momentarily could break existing code.

The following snippet:

```
import Result.Extra as Result
import Result.Extra

foo =
    Result.Extra.isOk (Ok ())
```

Is formatted as:

```
import Result.Extra as Result

foo =
    Result.Extra.isOk (Ok ())
```

The reference to `Result.Extra.isOk` results in a compilation error after the format.
I know that the code makes no sense, but apparently I had it in my codebase.

Version: `elm-format-0.18 0.6.1-alpha-48-ge91bc4e3`

avh4:
Yeah, can we enforce that rule (one import line per external module)?  It seems preferred, but are there cases where multiple `as`s are actually needed?  (Also, FYI, a future elm-format will be able to fix the qualified reference that now needs to use the `as` alias, but that could also lead to broken code in cases where there are multiple modules imported with the same `as`.)

stil4m:
I'm not sure how much information you have while rewriting the file, but I think the demonstrated case (two different qualifiers for a single module) is the only one that will break the code if its automatically resolved.

```
import Foo exposing (bar)
import Foo exposing (baz)

--would become
import Foo exposing (bar,baz)
```

```
import Foo exposing (bar)
import Foo exposing (..)

--would become
import Foo exposing (..)
```

```
import Foo as F exposing (bar)
import Foo as F exposing (baz)

--would become
import Foo as F exposing (bar,baz)
```

I think there is no good reasons to have two different `as` for the same module, but you are never sure. The major problems with this may arise with editor addons that support auto-completion, these might introduce duplicate imports which will be resolved by `elm-format` which in its turn might break the code (but I'm not sure about this either).

Wouldn't duplicate imports be a thing for the elm-compiler? It already gives warnings for unused imports, this might be a good warning as well.

avh4:
Cool, so the next release is going to follow the new release conventions and be an "experimental" release, so I'm going to err on the side of strictness for now.  If people run into this and don't end up liking it in real code, start a new issue and we can collect examples there.

PR #281, opened 2016-12-02
Update README.md for homebrew devel support

bmike:
elm-format 0.5.2-alpha is on a [development tap of homebrew](https://github.com/Homebrew/homebrew-devel-only/commit/23fc162559631de89927779f8a25789fbd3ca038) now


avh4:
Thanks for this, but it appears that the devel branch of homebrew has been migrated to the main branch now?  (I don't completely understand the situation.) 

I'd like to add official homebrew support in the future, but until that is done and someone is able to commit to maintaining the homebrew package, I'm not ready to update the readme to recommend homebrew.

PR #284, opened 2016-12-06
removed "export" from reserved

### LINKED ISSUE #283: Function with name "export" fails
I have a function named "export" like this:

```Elm
module BrokenExportKeyword exposing (..)

export : String
export = "export"
```

This compiles perfectly in Elm but elm-format does not like it:

```
-- SYNTAX PROBLEM --------------------------------------------- BrokenExport.elm

I ran into something unexpected when parsing your code!

3│ export : String
         ^
I am looking for one of the following things:

    a value definition

```

falti:
fixes #283


avh4:
Thanks!

PR #288, opened 2017-01-07
Add npm installer

rtfeldman:
To try it out: (this will install `elm-format 0.5.2-alpha` for Elm 0.18)

```
$ npm install -g /path/to/elm-format/package/npm
$ elm-format --version
```

Should work on the 64-bit versions of:

* macOS
* Linux
* Windows

## Notes

It installs `elm-format` versions based on the `version` specified in `package.json`, which is set up to include both the `elm-format` version and the Elm version.

For example, to explicitly get the latest version, you'd run `npm install -g elm-format@0.5.2-elm0.18.0` (although in the case of the absolute latest version you could always just do `npm install -g elm-format`)

This means when publishing a new version, you'd need to publish all the combinations as separate npm package releases.

## TODO

Add CI tests like what `elm-platform` uses, to verify all the combinations download and install successfully.

### DIFF
--- a/package/npm/install.js
+++ b/package/npm/install.js
@@ -1,21 +1,7 @@
 var binstall = require("binstall");
 var path = require("path");
 var fs = require("fs");
-var packageInfo = require(path.join(__dirname, "package.json"));
-
-var binVersion = packageInfo.version;
-
-// 'arm', 'ia32', or 'x64'.
-var arch = process.arch;
-
-// 'darwin', 'freebsd', 'linux', 'sunos' or 'win32'
-var operatingSystem = process.platform;
-
-var filename = operatingSystem + "-" + arch + ".tar.gz";
-var url = "https://dl.bintray.com/elmlang/elm-format/default/" +
-  binVersion +
-  "/" +
-  filename;
+var url = require(path.join(__dirname, "elmFormatDownload.js")).url();
 
 var binariesDir = path.join(__dirname, "binaries");
 var packageInfo = require(path.join(__dirname, "package.json"));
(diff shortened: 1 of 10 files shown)


rtfeldman:
@avh4 let me know what your npm username is and I'll add you as an owner to the `elm-format` package.

avh4:
https://www.npmjs.com/~avh4

avh4:
for version numbers, I think maybe we should only publish the elm-0.18 version via npm (it can be used for any other supported elm-version by passing `--elm-version=...`

and if we do want to publish other elm version versions, we could publish them as `elm-format-0.17@0.5.2`, etc.

What do you think of that?

rtfeldman:
> if we do want to publish other elm version versions, we could publish them as `elm-format-0.17@0.5.2`, etc.

Well, there's a slight risk of someone being a jerk and squatting on the package name before we can grab it. 😉 

I presume the upside would be that you could do `npm install elm-format@0.5.2`? It's possible to support that as an additional release on top of `elm-format@0.5.2-elm0.18.0` 😄

rtfeldman:
I guess there's also the issue that this would result in a lot of versions being listed when you looked at the module. Arguably that's a good thing though; would let you see all the supported combinations all in one place. 🤔

avh4:
Ah, okay.  Well I think we should start with only publishing elm-format that defaults `--elm-version` to the latest elm version, so `elm-format@5.2.0`.  And if there's later a good reason to publish multiple versions, we can decide between the version suffix and publishing multiple packages.

rtfeldman:
Wouldn't `elm-format@0.5.2` be more appropriate than `5.2.0`?

avh4:
yes, typo :)

rtfeldman:
@avh4 I made the package change, so I think the only thing remaining is to add some quick Travis tests to show that it installs and runs.

What I've done for these in the past is to write `.js` scripts to test them, but I see your Bash-fu is better than mine. How would you like to proceed?

package/npm/install.js
```diff
@@ -0,0 +1,35 @@
+var binstall = require("binstall");
+var path = require("path");
+var fs = require("fs");
+var packageInfo = require(path.join(__dirname, "package.json"));
+
+// Use major.minor.patch from version string - e.g. "1.2.3" from "1.2.3-alpha"
+var binVersion = packageInfo.version.match(/^(\d+\.\d+\.\d+).*$/)[1];
```
avh4: Does bintray not allow prerelease identifiers?

avh4:
(CI :x: because the prs/branches were deleted before CI finished)

PR #300, opened 2017-02-28
Clarify atom installation instructions

newswim:
This is a discussion-seeking PR. Following these instructions helped me get elm-format working but this process could benefit from other users' feedback.

I chose to rename references to the `elm-format` package since there's not a published package called `atom-elm-format` on apm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,8 @@ Find your editor in the table below.  The recommended plugin for each editor is 
   </tr>
   <tr>
     <td rowspan=2><a href="https://atom.io/">Atom</a></td>
-    <td>:trophy: <a href="https://atom.io/packages/elm-format">atom-elm-format</a></td>
-    <td>:white_check_mark: <a href="#atom-elm-format-installation">2 steps</a></td>
+    <td>:trophy: <a href="https://atom.io/packages/elm-format">elm-format</a></td>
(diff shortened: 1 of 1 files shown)


Rolograaf:
It maybe is not Windows friendly?

newswim:
Oh, I also didn't realize that you could just use `atom-beautify`, i read that as a typo.. I'll update this

README.md
```diff
@@ -159,30 +159,29 @@ If you can simplify or improve the installation instructions or add instructions
...
+2. Moving it to `usr/local/bin` will save some time later, although you can technically place it anywhere in your file system.
 
-### atom-beautify installation
-
+##### Install:
 1. Install elm-format
-1. Install atom-beautify
+2. Install atom-beautify
+```
+$ apm install elm-format
```
Glavin001: Atom-Beautify does not require `elm-format` from `apm` to be installed. It does require `elm-format` binary to be installed, which is already step 1.

avh4:
There seems to be some confusion here:  there are actually two different atom plugins, and you only need to choose one of them.  There hasn't been any activity here in a while, so if there's something we can change to clarify what needs to be done to get things installed, please open a new PR.  Thanks!

PR #316, opened 2017-03-12
JSON error reporting

avh4:
This will replace https://github.com/avh4/elm-format/pull/221


avh4:
@ento I rebased your PR, and I think I will own the branch from here, but I do still have some things to discuss:

  - I think the line and column fields shouldn't go in the JSON for now because they're misleading: the formatting error is not necessarily at line 1, column 1
  - I'm thinking that instead of having a `--format` flag, that simply using `--validate` will always give JSON output, since `--validate` is only meant for CI use, and that's when and only when you'll want this JSON

avh4:
I updated the reported message to include the elm-format and elm versions:
`message:"File is not formatted with elm-format-0.6.0-alpha --elm-version=0.18"`

ento:
Thanks!

> I think the line and column fields shouldn't go in the JSON for now because they're misleading: the formatting error is not necessarily at line 1, column 1

👍 Sounds good to me.

> I'm thinking that instead of having a --format flag, that simply using --validate will always give JSON output, since --validate is only meant for CI use, and that's when and only when you'll want this JSON

:art: Hmm, could editors calling elm-format without the --validate flag also benefit from a JSON output? or just showing human-readable outputs to the user as-is is fine now. Having fewer command line options (--validate always outputs JSON, period) sounds good in any case.

avh4:
JSON for editor plugins will come after v1.0.0, and I think error reporting for things that elm-format could just fix probably won't be useful for editors.

avh4:
(will merge after #318 is ✅ )

PR #317, opened 2017-03-13
Refactor using Free monads

avh4:


avh4:
cc @stoeffel @jwoudenberg @joneshf  does this look like a step in the right direction?

avh4:
(will merge after #318 is ✅ )

PR #318, opened 2017-03-14
--validate outputs a valid JSON array

avh4:
@ento this is the final step for https://github.com/avh4/elm-format/pull/316, and makes it output a valid JSON array with square brackets and commas.

(note: depends on #317, which I will merge, along with #316 if this PR looks good.)

### DIFF
--- a/src/ElmFormat/Execute.hs
+++ b/src/ElmFormat/Execute.hs
@@ -1,9 +1,10 @@
-module ElmFormat.Execute (forHuman, forMachine) where
+module ElmFormat.Execute (forHuman, forMachine, forMachineInit, forMachineDone) where
 
 {-| This module provides executors that can take streams of Operations and
 perform IO.
 -}
 
+import Control.Monad.State
 import ElmFormat.Operation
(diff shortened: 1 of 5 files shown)


src/ElmFormat/Execute.hs
```diff
@@ -22,9 +23,19 @@ forHuman operation =
...
 forMachine elmVersion operation =
     case operation of
-        InFileStore op -> FileStore.execute op
+        InFileStore op -> lift $ FileStore.execute op
         InInfoFormatter op -> Json.format elmVersion op
-        DeprecatedIO io -> io
+        DeprecatedIO io -> lift io
+
+
+forMachineInit :: (IO (), Bool)
```
ento: :art: It wasn't apparent to me that the state is for `printComma` (right?) -> A comment? I'm not sure of the volatility of the code around here so leaving it up to youl

avh4: Yeah, I was thinking about that too, but didn't have a clear answer, so decided to leave it TBD for now.

PR #341, opened 2017-04-01
Add gpg steps to Arch linux packaging guide.

mattjbray:
The Arch packages now validate the PGP signatures, so users need to import your key.


avh4:
Thanks for keeping this updated!

And FYI, elm-format-0.16 is no longer be supported starting in 0.6.0-alpha.

PR #385, opened 2017-07-18
Remove hiding and foreign from reserved words list

tac-tics:
I managed to break elm-format by naming one of my variables `hiding` earlier today. After tracking it down, I found that elm-format includes `hiding`, `foreign`, and `deriving` as reserved words (carry-overs from Haskell?) which are not keywords in (the current version of) Elm. 

I changed the keyword list to match [the keyword list in elm-compiler](https://github.com/elm-lang/elm-compiler/blob/master/src/Parse/Primitives.hs#L364-L375)

Cheers.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -7,6 +7,7 @@ Copyright (c) 2016, Max Goldstein
 Copyright (c) 2016, Daniel Bachler
 Copyright (c) 2016, Fedor Nezhivoi
 Copyright (c) 2013, John MacFarlane (Markdown parser forked from jgm/cheapskate)
+Copyright (c) 2017, Michael Maloney
 
 All rights reserved.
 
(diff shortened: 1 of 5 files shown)


avh4:
Thanks!  We should add tests to something like `tests/test-files/good/AllSyntax/0.18/NotKeywords.elm` for the words that are no longer keywords.

(Those all used to be keywords in Elm 0.16)

tac-tics:
Are the tests in the `good/` subdirectory meant to be fixedpoints of applying elm-format? If so, I can throw together a test file together.

avh4:
Looks great!

The last step is to add a commit to `/LICENSE`, adding yourself on a line after "John MacFarlane", confirming that you're licensing your contributions under the license there (BSD-3).

avh4:
Thanks!  Will be in 0.7.1 (no release date yet, but likely will be within a month)

PR #411, opened 2017-09-17
Update vscode status in README table

nloomans:
The Elm Language Support plugin for vscode supports format on save, see https://github.com/Krzysztof-Cieslak/vscode-elm/pull/62


avh4:
Thanks!

Issue #302, opened 2017-03-05
Eliminate compilation warnings


avh4:
I'd like to fix all haskell compilation warnings, and then turn on the `-Werror` flag (to make all warnings act as errors) for compilation.

avh4:
Note: There was a bug in stack < 1.4.0, causing ghc-options in the cabal file to not be honored.  (https://github.com/commercialhaskell/stack/issues/2289)

ahstro:
Would it be okay if I looked into this? I don't have a ton (see *close to none*) of Haskell experience, but I'm interested to learn, and I'm a sucker for getting rid of warnings.

avh4:
@ahstro sure!  The most common warning is that there are local variables that shadow outer scope variables, so coming up with good distinct names will probably be the bulk of the work.  One thing to note is that cabal won't rebuild files that haven't changed, so if you've already compiled everything, recompiling will only print out the warnings for files that need to be recompiled.

I'd suggest doing a PR for one file at a time, so there isn't a huge PR.

ahstro:
I saw that, I'll do my best trying to figure out good names. Guess that gives me a good reason to read and figure out some of the code, so feels like a good starting point.

Thanks for the tip on Cabal!

PR #420, opened 2017-10-25
Fix some compiler warnings

ahstro:
Fix compiler warnings of the following types:
 - `unused-imports`
 - `unused-top-binds`
 - `unused-matches`
 - `unused-do-binds `

Progress on #302 

@avh4 I know you suggested doing PRs for one file at a time, but this is what I had done so far, and it seemed easy enough to reason about, so I figured I'd make a PR with it before continuing. I can do a PR per file from now on or continue with one per compiler warning? (Or even build on this one, I think the total amount of compiler warnings was ~ 30, so the final diff wouldn't be that hard to consume, especially if doing so in commit-increments). It's up to you 😊

Also, let me know if something looks off or should be done differently. Like I said, not a long time Haskeller, and I couldn't find a style guide.

### DIFF
--- a/markdown/Cheapskate/Inlines.hs
+++ b/markdown/Cheapskate/Inlines.hs
@@ -15,7 +15,6 @@ import Prelude hiding (takeWhile)
 import Control.Applicative
 import Data.Monoid
 import Control.Monad
-import qualified Data.Map as M
 import Data.Text (Text)
 import qualified Data.Text as T
 import qualified Data.Set as Set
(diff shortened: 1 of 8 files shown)


avh4:
This one is great!  (And I didn't actually know about the `_name` syntax until now, so I learned something!)

I think the local variable name shadowing warning that I mentioned in the other issue is likely to be the only one that might need discussion.  Per-file PRs is not required, but the goal is to make sure the easy-to-approve stuff can get merged quickly, and doesn't get stuck in a PR with other stuff that needs some back and forth discussion before approving.

avh4:
(The commit-per-warning-type seems great, especially for these trivial-to-fix warnings.)

ahstro:
I'll try to split it up so the easily mergable changes are separate from the name shadowing. I imagine, as you say, that I won't necessarily get that one on the first go, and would like to be able to ask for and get feedback without bottlenecking other changes.

PR #424, opened 2017-11-05
[WIP] Replace Stack with Nix

### LINKED ISSUE #364: Replace stack with nix to support --split-objs
See https://github.com/Gabriel439/haskell-nix and https://stackoverflow.com/a/27970684/308930

ahstro:
Okay, so I don't know if this is good, or even still wanted, but I just spent a couple of hours playing with Nix to get the latest version of elm-format on my laptop and though I would share.

So far, this adds a Nix config files for elm-format + some dependencies that had to be explicitly versioned. I didn't go ahead and remove any Stack things yet (except for replacing the instructions in the README), because, well, frankly I don't quite get how that works (the `Dockerfile` scared me).

Closes #364

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,13 +299,11 @@ Please note that this project is released with a [Contributor Code of Conduct](C
 ### Building from source
 
 ```bash
-brew install haskell-stack
+curl https://nixos.org/nix/install | sh
 git clone https://github.com/avh4/elm-format.git
 cd elm-format
-stack setup
-stack build
-stack install
-~/.local/bin/elm-format-0.18 --help
+nix-env -i -f .
+elm-format-0.18 --help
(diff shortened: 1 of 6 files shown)


avh4:
Cool!  I realized I forgot to post info about an attempt I made.  Maybe compare to this branch: https://github.com/avh4/elm-format/tree/nix

Also, I realized nix can't easily be used to build binary distributions, because nix will link to libraries (like libc and the dynamic loader) that aren't in the normal locations.  But still it would be good to have an official release in the nix package repo.  (Documentation for how to release it that way can be put in `package/nix`)

ahstro:
Stopped using Nix, so not going to get around to this.

PR #428, opened 2017-11-17
Remove elm-format info for VSCode

JasonMFry:
There is no longer an `elm-format` extension for VSCode so I removed instructions on installing it and instead pointed the user back to using `elm`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ The default behavior of `elm-format`-approved plugins is to format Elm files on 
 
 ### VSCode Elm Format installation
 
-Note, Elm Format is no longer available for VS Code. Use [elm](visual-studio-code-installation) instead
+Note, Elm Format is no longer available for VS Code. Use [elm](#visual-studio-code-installation) instead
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -264,16 +264,7 @@ The default behavior of `elm-format`-approved plugins is to format Elm files on
...
-1. Install VSCode Elm Format
-
-    ```bash
-    ext install elm-format
-    ```
-
-1. You can run elm-format by using the `Elm: Format` command
-1. You can also run elm-format whenever you save the file by adding `formatOnSave: true` option to your [settings.json](https://code.visualstudio.com/docs/customization/userandworkspace) file.
-
+Note, Elm Format is no longer available for VS Code. Use [elm](visual-studio-code-installation) instead
```
avh4: The link needs to point to `#visual-studio-code-installation` to work properly.

JasonMFry:
In my opinion, yes we should remove it from the table. I'll do that now.

avh4:
Thanks for updating the info!

PR #433, opened 2017-12-09
Update info about atom-elm-format

anagrius:
You can format elm files using `Elm Format: File` now. No need to save.


anagrius:
@avh4 This should still be valid. The atom plugin does not need save in order to format a file.

PR #450, opened 2018-01-24
cleanup imports

alicebob:
Saves all warnings such as this one:

```
/.../elm-format/src/Messages/Formatter/HumanReadable.hs:10:1: warning: [-Wunused-imports]
    The import of ‘System.IO’ is redundant
      except perhaps to import instances from ‘System.IO’
    To import instances alone, use: import System.IO()
```

(There are still a lot of `-Wname-shadowing` warnings, but that's for another time)


avh4:
Contributes to https://github.com/avh4/elm-format/issues/302

avh4:
Thanks for chipping in!

PR #451, opened 2018-01-26
Fix almost all warnings

alicebob:
Most should be fine, but the last two commits might be too big a diff. Up to you what you do with this.

There are a few warnings left in a CLI module, but that one seems used only in tests (I'm not sure how to fix them).

### DIFF
--- a/parser/src/Parse/Type.hs
+++ b/parser/src/Parse/Type.hs
@@ -5,82 +5,82 @@ import Text.Parsec ((<|>), (<?>), char, many1, string, try, optionMaybe)
 
 import Parse.Helpers
 import qualified Reporting.Annotation as A
-import AST.V0_16
+import qualified AST.V0_16 as AST
 import Parse.IParser
 import Parse.Common
 import Data.Maybe (maybeToList)
 
 
-tvar :: IParser Type
+tvar :: IParser AST.Type
 tvar =
   addLocation
-    (TypeVariable <$> lowVar <?> "a type variable")
+    (AST.TypeVariable <$> lowVar <?> "a type variable")
 
 
-tuple :: IParser Type
+tuple :: IParser AST.Type
 tuple =
   addLocation $
   do  types <- parens'' (withEol expr)
       case types of
         Left comments ->
-            return $ UnitType comments
+            return $ AST.UnitType comments
         Right [] ->
-            return $ UnitType []
-        Right [Commented [] (t, Nothing) []] ->
+            return $ AST.UnitType []
+        Right [AST.Commented [] (t, Nothing) []] ->
             return $ A.drop t
-        Right [Commented pre (t, eol) post] ->
-            return $ TypeParens (Commented pre t (maybeToList (fmap LineComment eol) ++ post))
+        Right [AST.Commented pre (t, eol) post] ->
+            return $ AST.TypeParens (AST.Commented pre t (maybeToList (fmap AST.LineComment eol) ++ post))
         Right types' ->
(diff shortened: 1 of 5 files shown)


alicebob:
For #302

parser/src/Parse/Type.hs
```diff
@@ -5,82 +5,82 @@ import Text.Parsec ((<|>), (<?>), char, many1, string, try, optionMaybe)
 
 import Parse.Helpers
 import qualified Reporting.Annotation as A
-import AST.V0_16
+import qualified AST.V0_16 as AST
```
avh4: Did qualifying this help fix warnings, or was this just an additional cleanup?

src/Messages/Strings.hs
```diff
@@ -49,6 +49,8 @@ showErrorMessage OutputAndValidate =
 showErrorMessage (MustSpecifyVersionWithUpgrade elmVersion) =
     "I can only upgrade code to the latest Elm version.  To make sure I'm doing what you expect, you must also specify --elm-version=" ++ show elmVersion ++ " when you use --upgrade."
 
+showErrorMessage NoInputs =
+    "No inputs"
```
avh4: Why did this not lead to crashes before?  Is `NoInputs` actually used anywhere?

alicebob: src/ElmFormat.hs:
```
                    case (elmVersionResult, determineWhatToDoFromConfig config resolvedInputFiles) of
                        (_, Left NoInputs) ->
                            (handleParseResult $ Flags.showHelpText defaultVersion elmFormatVersion experimental)
                                >> exitFailure

                        (_, Left message) ->
                            exitWithError message
...
```

`exitWithError` uses `showErrorMessage` (which is what the warning is about). So the `NoInputs` case is effectively impossible because it is handled in a special way.

parser/src/Parse/Type.hs
alicebob: It's for these 4 warnings:

```
[53 of 61] Compiling Parse.Type       ( parser/src/Parse/Type.hs, .stack-work/dist/x86_64-linux-nopie/Cabal-1.24.2.0/build/Parse/Type.o )

.../elm-format/parser/src/Parse/Type.hs:41:13: warning: [-Wname-shadowing]
    This binding for ‘base’ shadows the existing binding
      imported from ‘AST.V0_16’ at parser/src/Parse/Type.hs:8:1-16

.../elm-format/parser/src/Parse/Type.hs:42:14: warning: [-Wname-shadowing]
    This binding for ‘fields’ shadows the existing binding
      imported from ‘AST.V0_16’ at parser/src/Parse/Type.hs:8:1-16

.../elm-format/parser/src/Parse/Type.hs:91:24: warning: [-Wname-shadowing]
    This binding for ‘first’ shadows the existing binding
      imported from ‘AST.V0_16’ at parser/src/Parse/Type.hs:8:1-16

.../elm-format/parser/src/Parse/Type.hs:91:31: warning: [-Wname-shadowing]
    This binding for ‘rest’ shadows the existing binding
      imported from ‘AST.V0_16’ at parser/src/Parse/Type.hs:8:1-16
```

Since thse names of the exported methods from `AST.V0_16` are so generic I choose to solve it with qualifying the import. Same reasoning for the other file in this PR where` AST.V0_16` is imported.

avh4:
Great, thanks!  I added 45c8e6b789d02ff66c00fcedec9f6d1fde00c616...101460c0a0141cebb08365ac205cc50f62a55ad7

Issue #72, opened 2015-12-07
Write haskell tests to test things in Main.hs


avh4:
The command line arguments and the recursive searching for .elm files in directories are currently tested in the `./tests/run-tests.sh` script.  Supposedly it is possible to write tests in haskell that test IO.  It would be great to get some examples of such haskell tests added so that we can replace the shell script tests for testing all the different command line arguments.

eeue56:
Got there a bit, since quickcheck has [monadic IO testing](https://hackage.haskell.org/package/QuickCheck-2.8.1/docs/Test-QuickCheck-Monadic.html) but didn't get anywhere useful.

avh4:
I'm currently putting tests around the `ElmFormat.Cli` module, but for integration tests, this looks interesting: https://hackage.haskell.org/package/tasty-program

PR #453, opened 2018-01-30
fold ElmFormat.Cli{,Test} into ElmFormat.Integration

robx:
This moves the two test cases that require `ElmFormat.Cli` more-or-less faithfully into `Integration`, which runs against the proper `ElmFormat.main` implementation. With minimal extensions to `TestWorld`. To deal with those `ElmFormat.Cli` compile warnings. Joint work with @alicebob.


robx:
(Not sure we want to keep the golden test call out to `diff`, used that for debugging, but it's sort of nice.)

alicebob:
For #302 again!

avh4:
This also relates to https://github.com/avh4/elm-format/issues/72

tests/Integration/CliTest.hs
```diff
@@ -16,12 +18,27 @@ tests =
             world
                 |> run "elm-format" [ "--help" ]
                 |> expectExit 0
+        , world
+            |> run "elm-format-xxx" []
+            |> TestWorld.goldenStdout "usage instructions"
+                "tests/Integration/data/usage.stdout"
```
avh4: I think you forgot to check this file in, or else forgot to change this path when you moved the file?

avh4:
Looks like there's one mistake here (see comment above).  Otherwise, this looks great!

robx:
Thanks for catching that! Indeed, I was moving files around later and tests were passing due to an untracked copy. Pushed a fix. Should I amend it into the buggy commit?

alicebob:
But Travis passed anyway?

avh4:
No need to amend.

Travis passed originally because apparently the behavior of golden is to create the file (and not error) if the file doesn't exist.

avh4:
Thanks!

Issue #378, opened 2017-06-12
npm install -g fails with `--ignore-scripts`


rtfeldman:
```
npm install -g elm-format@exp --ignore-scripts
```

```
elm-format --version
events.js:163
      throw er; // Unhandled 'error' event
      ^

Error: spawn /usr/local/lib/node_modules/elm-format/unpacked_bin/elm-format ENOENT
    at exports._errnoException (util.js:1050:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:193:32)
    at onErrorNT (internal/child_process.js:367:16)
    at _combinedTickCallback (internal/process/next_tick.js:80:11)
    at process._tickCallback (internal/process/next_tick.js:104:9)
    at Module.runMain (module.js:607:11)
    at run (bootstrap_node.js:427:7)
    at startup (bootstrap_node.js:151:9)
    at bootstrap_node.js:542:3
```

Potentially relevant info:

```
node --version
v7.10.0
```

```
npm --version
4.2.0
```

![screen shot 2017-06-12 at 10 13 27 am](https://user-images.githubusercontent.com/1094080/27046085-02ba6e30-4f58-11e7-9e72-6d86bf7f2e6b.png)

avh4:
This needs to be fixed in https://github.com/avh4/binwrap/issues/7

avh4:
This is fixed in the 0.8.0-rc3 release  `npm install -g elm-format@rc --ignore-scripts`

PR #463, opened 2018-03-01
Add a self-replacing binary for --ignore-scripts

rtfeldman:
Should fix https://github.com/avh4/elm-format/issues/378

### DIFF
--- a/package/npm/.gitignore
+++ b/package/npm/.gitignore
@@ -1,3 +1,2 @@
 /node_modules/
-/bin/
 /unpacked_bin/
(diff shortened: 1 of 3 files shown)


avh4:
Oh, interesting.  I think I'd like to make the binwrap pre-publish hook automatically create this wrapper so that it doesn't have to be manually created and updated.

Was jetpack blocked on this (or the binwrap PR) getting merged?

rtfeldman:
> Was jetpack blocked on this (or the binwrap PR) getting merged?

Nope! I'd like for us to get to a point where we can `npm install` everything with `--ignore-scripts=true` and this would remove a blocker to that, but it's not urgent.

rtfeldman:
Made a PR to `binwrap` to include instructions for doing this in the README: https://github.com/avh4/binwrap/pull/10

avh4:
An alternate approach was taken by upgrading to binwrap@0.2.0 in https://github.com/avh4/elm-format/pull/519

Issue #510, opened 2018-07-17
Add parentheses to import exposing clause if missing


andys8:
A common mistake I make quite often is forgetting parentheses for the types exposed when imorting data types. Here is an example:
```elm
-- Wrong
import Date exposing Date

-- Right
import Date exposing (Date)
```

The elm compiler will fail.
```shell
Failed to compile.

./src/Main.elm

-- SYNTAX PROBLEM ----------------------------------------------- ./src/Form.elm

I ran into something unexpected when parsing your code!

7| import Date exposing Date
                        ^
I am looking for one of the following things:

    a listing of values and types to expose, like (..)
    whitespace
```

How are the chances to having elm format fixing this automatically, which would awesome as hell? Can this be done in `elm-format` and would a pull request be merged?

avh4:
I agree, this sounds like a good idea.

For adding lenient parser features, I want to make sure we think through any possible syntax ambiguities that we might introduce.  The only one I can think of at the moment is

```
import Date exposing as as D
```

(which is not legal Elm, since Elm requires the `as` clause come before the `exposing` clause, but possibly "exposing before as" might be something elm-format might want to allow and correct in the future)

In any case, I think a PR for this would be good.  I expect it would simply involve a change to the parser that parses import lines, and for this type of feature, should add a new test case in `tests/test-files/transform/`

avh4:
I am tagging this for version 1.3.0, but if someone contributes a PR for this, I would be happy to include it in an earlier version.

avh4:
This was implemented in https://github.com/avh4/elm-format/pull/513  Thanks!

PR #513, opened 2018-07-21
Add parentheses to import exposing clause if missing

andys8:
Solves #510

### Changes

This PR adds the case that import statements with missing parentheses are allowed and will get parsed successfully, even they're not valid elm syntax. Because this is convenience, I intentionally made the decision not to include line breaks or comments, to keep things simple.

```elm
-- Wrong
import Date exposing Date

-- Right
import Date exposing (Date)
```

I'm trying to learn more about Haskell and am therefore open for hints (also "Allow edits from maintainers"). And a follow up question: Is there code formatting for the Haskell code base? How to use it?

### DIFF
--- a/parser/src/Parse/Module.hs
+++ b/parser/src/Parse/Module.hs
@@ -181,7 +181,11 @@ import' =
     exposing =
       do  preExposing <- try (whitespace <* reserved "exposing")
           postExposing <- whitespace
-          imports <- listing detailedListing
+          imports <-
+            choice
+              [ listing detailedListing
+              , listingWithoutParens
+              ]
           return (preExposing, (postExposing, imports))
 
 
@@ -202,6 +206,15 @@ listing explicit =
       sawNewline <- popNewlineContext
       _ <- char ')'
       return $ listing pre post sawNewline
+
+
+listingWithoutParens :: IParser (Var.Listing Module.DetailedListing)
+listingWithoutParens =
+  expecting "a listing of values and types to expose, but with missing parentheses" $
+  choice
+    [ (\_ -> (Var.OpenListing (Commented [] () []))) <$> string ".."
+    , (\x -> (Var.ExplicitListing (x [] []) False)) <$> detailedListing
+    ]
 
 
 commentedSet :: Ord a => IParser a -> IParser (Comments -> Comments -> Var.CommentedMap a ())
--- a/tests/test-files/transform/MissingImportParens.elm
+++ b/tests/test-files/transform/MissingImportParens.elm
@@ -1,0 +1,16 @@
+module Main exposing (x)
+
+import Import.As1 as ImportAs1
+import Import.As2 as ImportAs2 exposing A
+import ImportDotDot exposing ..
+import ImportName exposing A
+import ImportNames exposing A, B, x, y
+import ImportNamesLineBreak exposing
+    A, B
+import MergeImport exposing A, B
+import MergeImport exposing (Z)
+import OtherImport exposing (A(X, Y), b, z)
+
+
+x =
+    ()
--- a/LICENSE
+++ b/LICENSE
@@ -10,6 +10,7 @@ Copyright (c) 2013, John MacFarlane (Markdown parser forked from jgm/cheapskate)
 Copyright (c) 2017, Michael Maloney
 Copyright (c) 1999-2000, Daan Leijen. (part of Text.Parsec.Token)
 Copyright (c) 2007, Paolo Martini. (part of Text.Parsec.Token)
+Copyright (c) 2018, Andreas Scharf
 
 All rights reserved.
 
--- a/tests/run-tests.sh
+++ b/tests/run-tests.sh
@@ -395,6 +395,7 @@ checkTransformation 0.18 DocCommentCheapskateReferenceBug.elm
 checkTransformation 0.18 DocCommentAtDocs.elm
 checkTransformation 0.18 Sorting.elm
 checkTransformation 0.18 UnnecessaryParens.elm
+checkTransformation 0.18 MissingImportParens.elm
 checkUpgrade 0.18 Elm-0.18/PrimesBecomeUnderscores.elm
 checkUpgrade 0.18 Elm-0.18/RangesBecomeListRange.elm
 checkUpgrade 0.18 Elm-0.18/BackticksBecomeFunctionCalls.elm
--- a/tests/test-files/transform/MissingImportParens.formatted.elm
+++ b/tests/test-files/transform/MissingImportParens.formatted.elm
@@ -1,0 +1,14 @@
+module Main exposing (x)
+
+import Import.As1 as ImportAs1
+import Import.As2 as ImportAs2 exposing (A)
+import ImportDotDot exposing (..)
+import ImportName exposing (A)
+import ImportNames exposing (A, B, x, y)
+import ImportNamesLineBreak exposing (A, B)
+import MergeImport exposing (A, B, Z)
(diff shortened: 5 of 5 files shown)


tests/test-files/transform/MissingImportParens.elm
```diff
@@ -0,0 +1,14 @@
+module Main exposing (x)
+
+import Import.As1 as ImportAs1
+import Import.As2 as ImportAs2 exposing A
+import ImportDotDot exposing ..
+import ImportName exposing A
+import ImportNames exposing A, B, x, y
+import OtherImport1 exposing (A(X, Y), b, z)
+import OtherImport2 exposing (..)
+import OtherImport3
```
avh4: The "OtherImport*" lines are testing things that (hopefully!) should be tested in other files (in particular, somewhere in `tests/test-files/good/**`).  I'd probably leave them out here for that reason, though I could be convinced either way on this.

tests/test-files/transform/MissingImportParens.elm
```diff
@@ -0,0 +1,14 @@
+module Main exposing (x)
+
+import Import.As1 as ImportAs1
+import Import.As2 as ImportAs2 exposing A
+import ImportDotDot exposing ..
```
avh4: I wonder if we should really allow this one (the `..` without parens) to start with.  (I'm noting it's an explicit line of code that's added to allow this.)  This seems much less likely for people to actually type it this way.  I'd rather avoid adding this on speculation, but if you think this is actually common for people that will be benefiting from this feature, then I guess it's okay.  What do you think?

parser/src/Parse/Module.hs
```diff
@@ -181,7 +181,11 @@ import' =
     exposing =
       do  preExposing <- try (whitespace <* reserved "exposing")
           postExposing <- whitespace
-          imports <- listing detailedListing
+          imports <-
+            choice
+              [ listing detailedListing
+              , listingWithoutParens
```
avh4: Should we consider allowing missing parens for all listings, not just ones in the exposing clause of imports?  (I guess that's a question that can always be put off until later.)

parser/src/Parse/Module.hs
```diff
@@ -181,7 +181,11 @@ import' =
     exposing =
       do  preExposing <- try (whitespace <* reserved "exposing")
           postExposing <- whitespace
-          imports <- listing detailedListing
+          imports <-
+            choice
```
avh4: Since you asked for feedback:  The thing I'm thinking about when reviewing this code is what potential performance implication this `choice` has (which depends on whether the early choices allow for significant backtracking (with `try`) or not).  In this case, `listing` does `try (char '(')` so I think that will do a maximum of 1 character of backtracking, which should be fine.  (If there were the possibility of significant backtracking, then I'd want to look at ways to rewrite the choice to avoid the backtracking.)

In any case, I think it looks good for this particular change.

tests/test-files/transform/MissingImportParens.elm
```diff
@@ -0,0 +1,14 @@
...
+import ImportName exposing A
+import ImportNames exposing A, B, x, y
```
avh4: I think there are multiline cases to consider.

What will happen with 

```
import ModuleName exposing A,

f = ()
```
(I guess in this specific example, parsing would fail... but specifically, I'm wondering are there any cases where a typo might cause the new parsing branches to end up successfully consuming a bunch of code it shouldn't, and lead to an unexpected formatting result?)

Do we want to allow things like
```
import ModuleName exposing
   A, B
```

Either we should be conservative, and only this feature when everything's on a single line, or make sure we have test cases for the multiline variants we can think of, and think through whether there are any cases that could lead to a bad result.

avh4: Oh, you noted in the original post that you intentionally decided not to include line breaks... but `detailedListing` uses `commaSepBy1'`, which I think will consume comments, whitespace, and newlines immediately after a comma (maybe I'm wrong, though).  Did you try it to see what happens?

avh4:
Yeah, I haven't been using a code formatter for the haskell code.  I try to format it like Elm code (by hand) for new code, but I'm not really worried about it.

tests/test-files/transform/MissingImportParens.elm
andys8: `detailedListing` uses ` commaSep1'`.
https://github.com/avh4/elm-format/blob/master/parser/src/Parse/Helpers.hs#L246-L248
and `spaceySepBy1'`
https://github.com/avh4/elm-format/blob/master/parser/src/Parse/Helpers.hs#L236-L238
which uses `spaceySepBy1''`
https://github.com/avh4/elm-format/blob/master/parser/src/Parse/Helpers.hs#L214-L232

What's your idea to solve this?

andys8: The example above with the line breaks should and is not working. I'm adding a test file for bad tests. But to be honest, I can't foresee the edge cases and their consequences.

parser/src/Parse/Module.hs
andys8: Great, thanks for the info. I also thought about the order, and having the standard case at first, failing at the missing "(".

parser/src/Parse/Module.hs
andys8: So, `listing` is also used for type declarations and tuples? I'd be careful, in this case.

tests/test-files/transform/MissingImportParens.elm
andys8: There are arguments for both sides. There is an extra branch to support it and we're not sure, how likely it is for people to use. On the other side: Having a named value working, but ".." not, is inconsistent. Why is it like this and how should the user know?

Since it would work right now, I'd go and support it.

tests/test-files/transform/MissingImportParens.elm
andys8: You're right. My intention to have the good cases, was not to break them. While developing I had cases with a wrong parser, where the good case failed (because too much was parsed?). I'll reduce the standard cases, but I think it makes sense to have them here, to test the case, that there is a mix of different possibilities, and they they're working together in the same file. Also adding the merge case.

andys8:
Thanks a lot for the input! I rebased and added changes.

We could think about formatting the haskell code base and integrate tooling in the future, especially since this is a formatting tool's code base :)

tests/test-files/bad/MissingImportParens.elm
```diff
@@ -0,0 +1,4 @@
+module Main exposing (..)
+
+import ImportModuleWithoutParensWithLineBreak exposing
+   A, B
```
avh4: Elm files are invalid if they contain no definitions...  Does this test case still fail if you add `x = ()` (on its own line) at the end of the file?

andys8: You're right. This was the case and not the actual parsing to fail. It looks to my like parsing line breaks is working. Using the existing implementation I'm not sure what to expect regarding line breaks. Do you mind having a look into it? Is the current state of this PR working as we expect it?

andys8:
Sounds promising. Thanks @avh4

tests/test-files/transform/MissingImportParens.elm
avh4: Okay, playing with this a bit more, if there is a trailing comma, then it will consume the next token, but any subsequent elm definition must start at the beginning of a line, so if an extra token was consumed, then it won't be the beginning of the line anymore and will fail to parse.  So I think this seems safe. 👍

PR #515, opened 2018-07-28
Escape carriage returns with \r instead of hex code.

zenhack:
Right now elm-format transforms "\r" into "\x0D". This seems silly to me.


tests/test-files/good/AllSyntax/0.19/Literals.elm
```diff
@@ -77,9 +77,9 @@ multilineString =
 
 
 multilineString_keepsTrailingWhitespace =
-    """   3   
-  2  
- 1 
+    """   3
+  2
+ 1
```
zenhack: Gah, my editor strips out trailing whitespace by default. I'll fix and rebase.

zenhack:
I'll try to make a case:

* First, `\r` shows up all over the place in various file formats, protocols, etc. HTTP and SMTP both use it, and especially the former seems like something a web developer might benefit at least a little bit from knowing about.
* Re: utility for developers who do need to use it, In my case it came up in the context of generating CSV files, which use it as a line separator. Having the more readable to `\r` converted to a totally opaque hex code makes the source harder to understand; I'm never going to remember what the ascii code for carriage return is. A pretty printer should not actively hinder readability.
* If a developer isn't already familiar with the escape sequence, 30 seconds of googling should clear that up, so it doesn't a huge burden to learn.
* I will also argue that if `\r` is not to be used, a case should probably be made for just removing it from the language, rather than keeping it in the compiler and then having a standard tool ban it. I do appreciate the language designers' bias for leaving things out, and the fact that it seems to get smaller with each release. While I think it is defensible to say that `\r` is an unneeded feature, it seems like if it's is not worth having in the output of `elm-format`, that should imply it's not worth having in Elm.

zenhack:
I'll also add that developers not working on things where they need to know about carriage returns will probably be entirely unaffected by this change, as they'll just not see string literals with encoded carriage returns, regardless of how they're formatted.

billstclair:
In Elm 0.19.0, `"\x0D"` no longer parses. I get:

```
Backslashes always start escaped characters, but I do not recognize this one:

890|     "\x0D\n"
          ^
Maybe there is some typo?

Hint: Valid escape characters include:

    \n
    \r
    \t
    \"
    \'
    \\
    \u{03BB}
```

So for 0.19, you either have to leave `\r` alone, or change it to `\u{000D}`.

zenhack:
Yeah, elm-format changes it conditionally, based on the version of elm it's modifying (there's a command line flag to control this); you'll note my patch touches both.

zenhack:
I'm going to close this; it's clear there's no interest, and it's cluttering up my list of open prs.

billstclair:
I just discovered that my Emacs settings were using `--elm-version 0.18` . Changing it to 0.19 ensures that the code at least compiles. I'd still prefer to leave `"\r"` alone, but no biggy.

Thank you for `elm-format`. I use it every time I save an Elm buffer in Emacs.

zenhack:
For anyone interested: My Csv-generating module now defines a constant:

```elm
clrf : String
clrf =
    "\u{000D}\n"
```

To make things more readable elsewhere, so the ugly hex code is at least isolated.

PR #516, opened 2018-08-04
Simplify/correct vim installation instructions

jasoncarr0:
The existing instructions included an overly detailed overview of how to begin using the package manager vim-plug and use it to install vim-elm. Vim-plug is not a requirement, and including installation instructions for it is unnecessary and outright dangerous.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Find your editor in the table below.  The recommended plugin for each editor is 
   <tr>
     <td rowspan=1>Vim</td>
     <td>:trophy: <a href="https://github.com/ElmCast/elm-vim">elm-vim</a></td>
-    <td>:x: <a href="#elm-vim-installation">6 steps</a></td>
+    <td>:warning: <a href="#elm-vim-installation">3 steps</a></td>
     <td>:white_check_mark:</td>
     <td>⚠️ requires configuration</td>
     <td>:white_check_mark:</td>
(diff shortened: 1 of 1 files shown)


avh4:
I assumed experienced vim users would know whether or not they want to use vim-plug, and I wanted these instructions to be usable by beginners who don't know much about vim and without having to dig through multiple other sources of info to learn how to install it.  What should the steps be for users like that?  Is there an easy way for beginners to get vim-elm without using vim-plug?  What is dangerous about suggesting vim-plug to vim beginners?

jasoncarr0:
So, the reason that the steps should be shortened as requested, is that they will be simplified to some degree for users. Since these steps would be more direct, and less dense, the link going to the vim-elm page is highlighted. Following the link brings users to install instructions for the plugin, and an additional suggestion of vim-plug.

If we look at how it's phrased at the link:

"If you don't have a preferred installation method, I recommend installing vim-plug, and then simply add Plug 'elmcast/elm-vim' to your plugin section:"

This explains what vim-plug is, and how to install it with vim-plug, as well as linking directly to vim-plug's documentation to guide users in installing and using it without having to google anything.

Moreover, it does not make it an imperative. This is very important to keep users who just began using an alternative from being confused. If I'm a new user using Vundle for instance, my thoughts would be "Why do I need vim-plug?" or "Is this only available with vim-plug?". For some users, feeling they also have to install vim plug may make the steps more difficult, or leave them with a more complicated vimrc.

So it's my impression that including these steps makes the user experience for new users much worse, due to the redundancy (having extra, branching and nested steps), as well as the unexplained imperative. With this change, a new user will have links to appropriate documentation for installing, with no noise.

jasoncarr0:
I might be overstating the complexity a bit in there as I look back at it, but the summary is: It is certainly very dense, what's present. It far overstates the complexity of the installation for vim, which I think is overburdening. The imperative, which includes statements such as requiring a 'curl ... | sh' is dangerous to users.

jasoncarr0:
Any progress on a decision here?

avh4:
Thanks!  I want to keep the detailed steps for those who are newer to vim, since I think those are the folks most important to make sure they have clear instructions.  But I took your comments from this thread and tried to clarify things so as not to mislead advanced users: https://github.com/avh4/elm-format/pull/701/files  I hope that's better.

PR #518, opened 2018-08-18
Merge the WIP JSON AST work into master (it's not visible to users yet, though)

avh4:


avh4:
I'm guessing Appveyor failure is due to EOL characters

PR #528, opened 2018-08-24
Add instructions for Vim Ex command setup

VadimBrodsky:
I didn't want to install the Vim plugin, but I want to run `elm-format` from Vim. This is a small Ex mode command that will run `elm-format` on a buffer save of an `.elm` file.

The first example is more verbose and adds an explicit `:ElmFormat` ex command as well as a leader key binding.

The second example just runs `elm-format` on save without defining an explicit Ex command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,22 @@ The default behavior of `elm-format`-approved plugins is to format Elm files on 
   ```
   let g:elm_format_autosave = 1
   ```
+  
+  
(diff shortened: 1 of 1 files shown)


avh4:
Thanks!  I'm hesitant to merge this, as I want to try to keep the install options as simple as possible for folks who are not experts with their particular editor.  But I'll leave this issue open for reference by anyone who might be looking for this.

Issue #529, opened 2018-08-25
Easier to package upstream


domenkozar:
Hi, as I'm packaging this for Nix, a few notes:

- sync up to latest LTS (remove indents-0.4.0.0 dependency)
- Setup.hs depends on git invocations, but we use tarball source that does not have `.git`

avh4:
I would be happy to accept a PR to upgrade to the latest stack LTS snapshot.

For the Setup.hs use of git, the patch in the existing nix formula seemed to work well: https://github.com/NixOS/nixpkgs/blob/2c50340575a4b91bd983894d294e5619ae4aee96/pkgs/development/compilers/elm/packages/elm-format.nix#L34-L37

If you have a better suggestion for how to set up the project to automatically have a correct version number for all builds (including dev builds) that's different from calling git in Setup.hs, I'd be happy to hear about.  The reason for the git call in the first place is that I wanted to make sure that binaries build from any commit would include an appropriate version number so that any elm-format binary could be easily identified.

avh4:
Also, can the updating of the nix package be automated in some way?  If you want to contribute a PR to this repo to add a `package/nix` folder with a script or instructions that I can follow whenever I do a new elm-format release, I'd be happy to help kick off the updates of the nix package.

basile-henry:
One option would be to use something like the [`gitrev`](https://hackage.haskell.org/package/gitrev-1.3.1/docs/Development-GitRev.html) package to get the git commit at compile time? Then the `nix` derivation just needs to add `git` as a build dependency.

I was curious how `gitrev` is used across packages in `nixpkgs`, so I looked at the [hackage reverse dependencies for `gitrev`](https://packdeps.haskellers.com/reverse/gitrev) and picked the first package: `Agda`. It looks like nothing special is done to depend on `git` for `Agda` (neither `Agda` or `gitrev` depend on `git`) in `nixpkgs`.

If `gitrev` doesn't have access to `git` it returns an `UNKNOWN` revision. This is how `Agda` deals with `UNKNOWN` values: https://github.com/agda/agda/blob/d7e3a5883c918abaebe28f9d05ddc74c9ac88a6c/src/full/Agda/VersionCommit.hs#L17

So if you get `Agda` from `nixpkgs` you don't have access to the `git` revision it was built with. :confused: 

[...]

PR #559, opened 2018-10-03
Automate updating the nix package

avh4:
Includes #530, and addresses #529.

cc @domenkozar @basile-henry

I've got this working to create an update branch against nixpkgs, and test it.

However, one remaining problem, if you build this (with `cd package/nix && ./build.sh`) and then run the compiled binary with `nixpkgs/result/bin/elm-format`, the compiled binary has an empty string for the version number, so I'm thinking maybe the postPatch sed lines aren't quite right, but I have to learn how to get nix-build to save the working directory for me so I can debug.

### DIFF
--- a/package/nix/generate_derivation.sh
+++ b/package/nix/generate_derivation.sh
@@ -1,37 +1,30 @@
 #! /usr/bin/env nix-shell
-#! nix-shell -i bash -p git haskellPackages.cabal2nix nix-prefetch-git jq
+#! nix-shell -i bash -p git cabal2nix nix-prefetch-git jq
 
 # This script generates the nix derivation for elm-format intended for nixpkgs:
 # https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/compilers/elm/packages/elm-format.nix
 
-# To use this script, update the FLAVOR to the most recent flavor of elm-format
-# and then run:
+# To use this script, run:
 #
-# $ ./generate_derivation.sh > elm-format.sh
+# $ ./generate_derivation.sh > elm-format.nix
 #
 # This might take a bit of time if the dependencies are not already in the nix
 # store. If you already have all the dependencies installed, feel free to remove
 # them from the shebang to speed up the process.
 
-FLAVOR="0.19"
+set -exo pipefail
 
 REV="$(git rev-parse HEAD)"
-VERSION="$(git describe --abbrev=8)"
+VERSION="$(git describe --abbrev=8 "$REV")"
 ROOTDIR="$(git rev-parse --show-toplevel)"
-SHA="$(nix-prefetch-git --url "$ROOTDIR" --rev "$REV" --quiet --no-deepClone | jq .sha256)"
+SHA="$(nix-prefetch-git --url "$ROOTDIR" --rev "$REV" --quiet --no-deepClone | jq --raw-output .sha256)"
 
 PATCH=$(cat <<END
   src = fetchgit {
     url = "http://github.com/avh4/elm-format";
-    sha256 = $SHA;
+    sha256 = "$SHA";
     rev = "$REV";
   };
-
-  doHaddock = false;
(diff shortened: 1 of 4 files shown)


package/nix/generate_derivation.sh
```diff
@@ -1,20 +1,17 @@
 #! /usr/bin/env nix-shell
-#! nix-shell -i bash -p git haskellPackages.cabal2nix nix-prefetch-git jq
+#! nix-shell -i bash -p git cabal2nix nix-prefetch-git jq
```
avh4: `haskellPackages.cabal2nix` seemed to have a ton (~15GB) of dependencies that it pulled in, but switching to just `cabal2nix` apparently doesn't need them 🤷‍♀️

package/nix/generate_derivation.sh
```diff
@@ -1,20 +1,17 @@
...
 #
-# $ ./generate_derivation.sh > elm-format.sh
+# $ ./generate_derivation.sh > elm-format.nix
 #
 # This might take a bit of time if the dependencies are not already in the nix
 # store. If you already have all the dependencies installed, feel free to remove
 # them from the shebang to speed up the process.
 
-FLAVOR="0.19"
-
```
avh4: No more flavors as of elm-format 0.8.1 🎉

package/nix/generate_derivation.sh
```diff
@@ -29,9 +26,7 @@ PATCH=$(cat <<END
 
   doHaddock = false;
   jailbreak = true;
-  postInstall = ''
-    ln -s \$out/bin/elm-format-$FLAVOR \$out/bin/elm-format
-  '';
+  doCheck = false;
```
avh4: nixpkgs doesn't provide tasty-quickcheck anymore, so turning off checks was the simplest fix.

basile-henry:
Since I made PR #530, the `elm-format` derivation in `nixpkgs` has been changed: https://github.com/NixOS/nixpkgs/commit/7c542d0c86ad5958d2772026cf48939c3aee2aa7

This commit overrides the `patchPhase` entirely so the one we're generating with the `generate_derivation.sh` script isn't being used. That version of the patch sets the git hash as an empty string. If we want to keep the git hash as part of the binary we need to remove that `patchPhase` from `elm/default.nix` and only set it from `elm-format.nix`.

Maybe there are other ways to do this, so I'll wait to hear if @domenkozar has any better suggestions! :smile:

domenkozar:
I'd still prefer CPP over `sed`ing the source, it's less error prone.

So something like telling the build that it's a stable release, which would use Paths_elm_format.version

But any automation is a good step :)

avh4:
> the elm-format derivation in nixpkgs has been changed: NixOS/nixpkgs@7c542d0

Ah thanks, I didn't expect that there were things outside of `elm-format.nix` that were relevant!

> I'd still prefer CPP

What does "CPP" mean?

basile-henry:
> What does "CPP" mean?

It's the C pre-processor. There is a GHC extension that enables it and it is used fairly often (a bit too often) in Haskell for compile time configuration.

It would go something like this:

```haskell
{-# LANGUAGE CPP #-}

...

gitHash :: String
#ifdef GIT_HASH
gitHash = GIT_HASH
#else
gitHash = ""
#endif
```

And now we're putting the burden of providing `GIT_HASH` onto the build tools without having to modify the source code. 😄

domenkozar:
Exactly, you could invoke git and try to parse revision, otherwise fall back to version string from cabal (using something like https://github.com/cachix/cachix/blob/master/cachix/src/Cachix/Client/Env.hs#L53)

avh4:
Okay, cool!   I'll split the cpp stuff into a separate issue to look at later.  So if the updated https://github.com/NixOS/nixpkgs/pull/47651 gets merged, I'll merge this PR.

PR #564, opened 2018-10-07
Fix #544, Fix formatting exports with commented lines

### LINKED ISSUE #544: Extra space inserted in exposing list with comment
As discussed on Slack, elm-format adds two spaces between the comma and the function name in `exposing ()` if there is a comment below a function. 
```
module Test exposing
    ( test1
    ,  test2
       -- this comment causes the line above to be indented one space too far

    )
```
elm-format-0.19 0.8.0

mdevlamynck:
Fixes #544.

I'm not used to haskell so this PR might need some code style or code organization changes.

This PR aligns the formatting behavior of exports list on the formatting behavior of imports list.

Before:
```elm
module Exports exposing
    (  -- comment
       a
       -- comment
       -- comment

    ,  b
       -- comment

    )

import Imports
    exposing
        ( -- comment
          a
          -- comment
          -- comment
        , b
          -- comment
        )
```
After:
```elm
module Exports exposing
    ( -- comment
      a
      -- comment
      -- comment
    , b
      -- comment
    )

import Imports
    exposing
        ( -- comment
          a
          -- comment
          -- comment
        , b
          -- comment
        )
```

# Details:

We have a `[[AST.Commented Value]]` containing the exported values. It's a list of lists to handle cases like this one:
```elm
module Exports exposing
    ( a, b
    , c, d
    )
```
The lines are formatted with a call to `ElmStructure.group'` and the content of the line with a call to `ElmStructure.group`. To make things more apparent the examples bellow will use delimiters for both `group` calls (the outer already uses `"("` and `")"`, the inner will uses `"<"` and `">"` instead of empty strings).
The example above becomes:
```elm
module Exports exposing
    ( <a, b>
    , <c, d>
    )
```
The handling of commented lines (call to `formatCommented`) is on the inner list. With a comment the example becomes:
```elm
module Exports exposing
    ( < a, b
        -- comment
        >
    , <c, d>
    )
```
Since the comment forces a line break, the content of the first inner list is not a single line anymore so `group` switches to the multi line formatting.

The fix consists in extracting the comments to be able to call `formatCommented` on the already formatted line:
```elm
module Exports exposing
    ( <a, b>
      -- comment
    , <c, d>
    )
```
To do that we need to convert our `[[AST.Commented Value]]` to a `[AST.Commented [Value]]` merging the `pre` and `post` comments along the way (done by `mergeCommented`). Since a comment will always create a new line, doing this should not mess up the ordering of AST items. We can now format the line as usual with `ElmStructure.group` and `formatVarValue` and finally format comments for the whole thing with `formatCommented`.


avh4:
Good start, and I thanks for the clear explanation!

I'm worried that this might move comment around in certain cases, since it's pulling up the comments from each group.  I'm wondering what this code will do to inputs like:

```elm
module X exposing
    ( {- A -} A, {- B -} B
    , {- x -} x
    )
```

or 

```elm
module X exposing
    ( -- A1
      A
      -- A2
    , -- B1
      B
      -- B2
    )

{-|
@docs A, B
-}
```

mdevlamynck:
My bad, you're right on the second example it moves the comments to the end.

```elm
module X exposing
    ( {- A -} A, {- B -} B
    , {- x -} x
    )

-- Becomes

module X exposing
    ( {- A -} A
    , {- B -} B
    , {- x -} x
    )
```
```elm
module X exposing
    ( -- A1
      A
      -- A2
    , -- B1
      B
      -- B2
    )

{-|
@docs A, B
-}

-- Becomes

module X exposing
    ( A
    , B
    -- A1
    -- A2
    -- B1
    -- B2
    )

{-|

@docs A, B

-}
```
But weirdly it seems to be caused by the `@docs`, without it everything is fine.
```elm
module X exposing
    ( -- A1
      A
      -- A2
    , -- B1
      B
      -- B2
    )

-- Becomes

module X exposing
    ( -- A1
      A
      -- A2
    , -- B1
      B
      -- B2
    )
```
It also seems to be the reason the test mentioned in the previous comment does not fail (adding the `@docs` in the test makes it fail).

Do you know what the presence of `@docs` changes in the AST to provoke such different behavior?

mdevlamynck:
Actually this PR isn't introducing these weird behaviors. I tried your two examples with `elm-format@0.8.1` and I have the same results as reported in the previous comment. It seems to be caused by some transformations on the AST prior to calling `formatModuleLine`. Should this be handled in a separate issue?

avh4:
Ah, okay, I guess that's a separate bug, then.

The `@docs` are used to sort and group the exposed values.  Anything not mentioned in the docs is sorted alphabetically, but won't be grouped.  So using the `@docs` is the only way to get output that tries to put the exposed values into groups on the same line.

mdevlamynck:
Oh I see, so the first example should be considered ok (no `@docs` so each item is on it's own line). I'll create an issue for the second case.

I think once the `@docs` issue is solved, this PR will exhibit the comments reordering problem you mentioned. I'll try to fix it.

avh4:
This was rebased and merged in https://github.com/avh4/elm-format/pull/756  Thanks!

PR #565, opened 2018-10-07
Fix #454, Remove (..) from types in @docs

### LINKED ISSUE #454: Remove illegal (..) in @docs lines
As described in https://github.com/elm-lang/elm-package/issues/291, the following is allowed by elm-make, but causes the package website to break:

```elm
{-|
@docs SomeType(..)
-}
```

elm-format could implement a workaround for this by formatting the above as

```elm
{-|
@docs SomeType
-}
```

mdevlamynck:
Fixes #454.

This adds a cleanup pass before rendering the module docs content. It removes `(..)` if any in the `Markdown.ElmDocs`.

I'm not sure if this is the best place to put this pass but I figured if it's done in the formatting function it will not break other stuff and will not miss any `(..)` it's supposed to remove.

I'm not very familiar with haskell so my `Text.replace (Text.pack "(..)") (Text.pack "")` thing might not be the best way of doing this.

### DIFF
--- a/src/ElmFormat/Render/Box.hs
+++ b/src/ElmFormat/Render/Box.hs
@@ -700,7 +700,18 @@ formatModuleDocs elmVersion importInfo blocks =
 
         content :: String
         content =
-            ElmFormat.Render.Markdown.formatMarkdown (fmap format . parse) blocks
+            ElmFormat.Render.Markdown.formatMarkdown (fmap format . parse) $ fmap cleanBlock blocks
+
+        cleanBlock :: Markdown.Block -> Markdown.Block
+        cleanBlock block =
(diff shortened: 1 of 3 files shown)


avh4:
This looks good, thanks!

A possible improvement would be to instead put the code in the markdown parser so that the invalid characters are stripped out of the AST representation.  But given that the markdown parser is forked from another project and is quite confusing, I'm not sure it would be easy to do that way.

avh4:
@mdevlamynck oh, can you also add a line for yourself at the bottom of the list of authors in LICENSE, and make a PR indicating that you are licensing your contributions to this project under that BSD 3-Clause license

PR #574, opened 2018-11-05
Reference prettier-plugin-elm from README.md

kachkaev:
Just thought that a link to [`prettier-plugin-elm`](https://github.com/gicentre/prettier-plugin-elm) could be potentially useful to newcomers (especially given that Prettier is getting very popular amongst people who write JavaScript and Markdown).

Congratulations on 1,000 GitHub stars @avh4! 🌟

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ Find your editor in the table below.  The recommended plugin for each editor is 
 </table>
 
 Alternatively, you can utilize elm-format via a [Prettier plugin](https://github.com/gicentre/prettier-plugin-elm) that wraps it.
-This can be convenient if your project contains files in other languages like JavaScript, Markdown, etc.
+This can be convenient if your project contains files in other languages and you want to format all of them using a single API.
 [Prettier](https://prettier.io/) has its own integrations into editors and can also be used from a command line (including CI pipelines).
 
 ## Detailed instructions
--- a/README.md
+++ b/README.md
@@ -151,6 +151,9 @@ Find your editor in the table below.  The recommended plugin for each editor is 
   </tr>
 </table>
 
+Alternatively, you can utilize elm-format via a [Prettier plugin](https://github.com/gicentre/prettier-plugin-elm) that wraps it.
+This can be convenient if your project contains files in other languages like JavaScript, Markdown, etc.
+[Prettier](https://prettier.io/) has its own integrations into editors and can also be used from a command line (including CI pipelines).
 
 ## Detailed instructions
 


kachkaev:
ping @avh4 👋

kachkaev:
🏓

kachkaev:
@avh4 it’d be cool if you could share your thoughts on this addition to README. If you don’t like `prettier-plugin-elm` and don’t think we should reference it in README – that’s fine, we'll just close this PR. Otherwise, I'm happy to clean it if needed to make the merging possible.

avh4:
Thanks for the ping!

My conflicting goals for the README are to keep it as simple as possible while also giving the info that newcomers would be looking for and not sure how else to find.  Including information about editor plugins is included because it's meant to be the primary way that folks start out using elm-format.

I think I don't want to push folks in the direction of using prettier as the default option because it adds an extra tool they'd need to learn about and configure.  But if a lot of people tend to be using it already and would naturally wonder about whether elm-format is supported by it, then it could be worth including.  I also wouldn't want to recommend prettier over other similar options.

I'm thinking maybe a good way to do this is to just add a list of other tools that have elm-format integration, that way it's it can be seen by people who are familiar with those tools and also is clear that you can skip it if you're not already using those other tools.  How does that sound?

kachkaev:
I like the idea of a new section, such as _Integration into tools_. Do we know of any alternatives for Prettier which would also support `elm-format`? Asking so that we could start with a list with more than one list item.

avh4:
Yeah, I don't know of any similar formatting tools,  but I was thinking built tools like webpack, etc might have plugins for code formatters?

I'm not really up-to-date on those sorts of tools.  If we don't happen to know of any other such plugins, I'd be fine with having prettier-plugin-elm be the only one in that new section for now.

kachkaev:
I’m also fine with Prettier being the only item in the new section to start with. If you have an image of how that section would be called and structured, WDYT of closing this PR and updating README separately?

I’m happy to continue working on this PR, I just think that it might take you more time to review my text than to write your own. I’m easy about not being added to project contributors if the PR is closed without merging.

avh4:
@kachkaev I made https://github.com/avh4/elm-format/pull/729/files?short_path=b335630#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5  how does that look?

kachkaev:
Looks perfect! Happy for you to close this PR after merging `non-editor-tools` into `main`!

avh4:
Replaced by https://github.com/avh4/elm-format/pull/729

PR #586, opened 2018-12-17
use face emojis instead of status indicators in README integrations table

cameron:
The metric behind the status emojis in the installation steps column seems to be "level of effort required", but the status icons used imply a much harder pass-fail spectrum, which initially made me think the elm-format plugin doesn't work w/emacs' elm-mode. Faces don't imply any breaking failures 😃


avh4:
Thanks, I like this idea!  Though I think there are too many levels of faces, and it might not be clear which faces are better than others, and they're also more difficult to distinguish at a glance.  How about leaving the green checkmarks, and maybe only changing the red ❌'s to something else?

cameron:
Yeah that makes sense. I replaced only :x: with 😞, and now think it's probably best just to use ⚠️ . LMK!

PR #488, opened 2018-05-28
Change JetBrains plugin

klazuka:
The JetBrains plugin previously developed by @durkiewicz has been inactive for about a year now. And he [intends to remove it](https://github.com/durkiewicz/elm-plugin/issues/22#issuecomment-352651205) from the JetBrains plugin repository in favor of the new plugin that I wrote.

PR #600, opened 2019-03-28
Update JetBrains installation guide

Herteby:
The Elm Plugin now supports elm-format, so the File Watcher is no longer necessary, and it should work without configuration.


avh4:
The information on the linked plugin page says:

> Formatting is currently not a feature of the plugin, but elm-format can be used instead. If you don't know how to configure IntelliJ to use elm-format you can find some information here

is that info incorrect?   Does installing the plugin actually enable elm-format?

avh4:
Ah, I see: there's another PR that has a more complete update: https://github.com/avh4/elm-format/pull/488/files

Herteby:
@avh4 You have to install elm-format yourself, but the plugin does have an integration for it now (it didn't use to) and when you first open an Elm project you get asked to fill in the required details.

Screenshot:
image

PR #603, opened 2019-06-05
add a few more cases in the JSON output

BrianHicks:
just expanding on the work we did a while back. Added floats, chars, strings, and expanded the rest of the values for TODOs.

### DIFF
--- a/parser/src/AST/Json.hs
+++ b/parser/src/AST/Json.hs
@@ -185,6 +185,9 @@ instance ToJSON Expr where
           VarExpr (OpRef (SymbolIdentifier sym)) ->
             variableReference region sym
 
+          VarExpr _ ->
+              JSString $ toJSString "TODO: VarExpr"
+
           App expr args _ ->
               makeObj
                   [ type_ "FunctionApplication"
@@ -334,9 +337,6 @@ instance ToJSON Expr where
                         branches
                     )
                   ]
-
-          VarExpr _ ->
-              JSString $ toJSString "TODO: VarExpr"
 
           Range _ _ _ ->
               JSString $ toJSString "TODO: Range"
--- a/parser/src/AST/Json.hs
+++ b/parser/src/AST/Json.hs
@@ -132,6 +132,17 @@ instance ToJSON Expr where
                     )
                   ]
 
+          AST.Expression.Literal (FloatNum value repr) ->
+              makeObj
+                  [ type_ "FloatLiteral"
+                  , ("value", JSRational False $ toRational value)
+                  , ("display"
+                    , makeObj
+                        [ ("representation", showJSON repr)
+                        ]
+                    )
+                  ]
+
           AST.Expression.Literal (Boolean value) ->
             makeObj
                 [ type_ "ExternalReference"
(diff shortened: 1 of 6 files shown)


avh4:
Rebased (force push) to update the reference json in each commit.  (Helps me see what effect the code at each commit has.)

parser/src/AST/Json.hs
```diff
@@ -132,6 +132,17 @@ instance ToJSON Expr where
                     )
                   ]
 
+          AST.Expression.Literal (FloatNum value repr) ->
+              makeObj
+                  [ type_ "Floatiteral"
```
avh4: typo! should be "FloatLiteral"

parser/src/AST/Json.hs
```diff
@@ -151,6 +151,22 @@ instance ToJSON Expr where
                 , sourceLocation region
                 ]
 
+          AST.Expression.Literal (Chr chr) ->
+              makeObj
+                  [ type_ "ChrLiteral"
```
avh4: I'd prefer "CharLiteral", using the Elm type name, since I expect the end-users of this AST to be mostly familiar with Elm.

parser/src/AST/Json.hs
```diff
@@ -151,6 +151,22 @@ instance ToJSON Expr where
...
+                  , ("module", JSString $ toJSString "Char")
+                  , ("value", JSString $ toJSString [chr])
+                  , sourceLocation region
+                  ]
+
+          AST.Expression.Literal (Str str _) ->
+              makeObj
+                  [ type_ "StrLiteral"
```
avh4: I'd prefer "StringLiteral", using the Elm type name, since I expect the end-users of this AST to be mostly familiar with Elm.

tests/test-files/good/AllSyntax/0.18/Literals.json
```diff
@@ -169,7 +183,21 @@
...
+                        "col": 22,
+                        "line": 34
+                    },
+                    "start": {
+                        "col": 5,
+                        "line": 34
+                    }
+                },
+                "type": "StrLiteral",
+                "value": "a b'`<>{}/\u06a5\ud83d\ude00\u3041\u2f83\ud835\udfdf"
```
avh4: Do things work correctly when something that becomes an escaped unicode character is immediately followed by a hex digit?  Even if so, it looks like an example of that happening is missing from the `Literals.elm` test file

parser/src/AST/Json.hs
```diff
@@ -335,8 +335,17 @@ instance ToJSON Expr where
...
+              JSString $ toJSString "TODO: VarExpr"
+
+          Range _ _ _ ->
+              JSString $ toJSString "TODO: Range"
+
+          AccessFunction _ ->
+              JSString $ toJSString "TODO: AccessFunction"
+
+          GLShader _ ->
+              JSString $ toJSString "TODO: GLShader"
```
avh4: Now's probably a good moment to add a json test (probably just add the json transofmration check to an existing elm fixture test)  for these new things.

avh4:
thanks for pushing this forward!   I guess I'm planning on addressing my own comments later today.

avh4:
🤔 also, now I'm wondering if it's better for all literals to be a `type: "Literal"` node with the values being a union type of `StringValue | IntValue | CharValue` etc

tests/test-files/good/AllSyntax/0.18/Literals.json
avh4: Apparently it works...  though the choice in escaping is a little weird imo.  This literal string `😀0😀9😀a😀F` results in the following json respresentation: `"\ud83d\ude000\ud83d\ude009\ud83d\ude00a\ud83d\ude00F"`

But I guess the details of the escaping don't really matter, since the consumers of the API can be expected to be using a compliant JSON parser.

Issue #607, opened 2019-06-21
Try formatting the haskell code with ormolu


avh4:
https://github.com/tweag/ormolu

NOTE: please coordinate with any other in-progress branches to avoid painful merge conflicts

andys8:
Oh, interesting. I heard about `ormolu` recently at Zurihac where people were working on it. Are your experiences better with it than e.g. `brittany` or `floskell`?

avh4:
I haven't tried any of them yet.  So I guess this issue is for comparing them.

I think I'd prefer one that is aligned with elm-format if possible.  I'd weigh these criteria in this order::
- no configuration
- is similar in philosophy to elm-format
- is similar in output style to elm-format

PR #626, opened 2019-08-04
Format repository with brittany 0.12.0

andys8:
This is the current repository formatted with [`brittany`](https://github.com/lspitzner/brittany) with default settings (no configuration). @avh4 If you like it, feel free to merge, otherwise reject it.

https://hexagoxel.de/postsforpublish/posts/2017-10-28-brittany-rationale.html

Related to #607

### DIFF
--- a/BuildFromSource.hs
+++ b/BuildFromSource.hs
@@ -34,77 +34,82 @@ and testing interactions between projects.
 -}
 module Main where
 
-import qualified Data.List          as List
-import qualified Data.Map           as Map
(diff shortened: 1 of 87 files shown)


avh4:
Thanks for trying this out!

I'm gonna stick with manual formatting for now, to try to still maintain at least a bit of mergability with the elm-compiler code.  Plus it seems like there hasn't been a clear consensus yet on which formatter haskell folks prefer.

I imagine this would be easy to recreate again if necessary.  Further comments can go to -> https://github.com/avh4/elm-format/issues/607

PR #632, opened 2019-08-09
Add FreeBSD installation instructions

evilham:
Current pre-compiled version is `0.8.1`, but there is a change pending review for `0.8.2`
See: https://reviews.freebsd.org/D21202

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ or download the version appropriate for your OS from the [release page](https://
 unzip it,
 and place `elm-format` or `elm-format.exe` (windows) on your `PATH`.
 
+On FreeBSD you can `pkg install hs-elm-format` or use [devel/elm-format](https://www.freshports.org/devel/elm-format/), then execute `elm-format` normally.
+
 You must run `elm-format` from the directory that contains your `elm.json` (for Elm 0.19) or `elm-package.json` (for Elm 0.18),
 or else you must pass the appropriate `--elm-version=0.19`/`--elm-version=0.18` command line argument.
 


README.md
```diff
@@ -45,6 +45,8 @@ or download the version appropriate for your OS from the [release page](https://
 unzip it,
 and place `elm-format` or `elm-format.exe` (windows) on your `PATH`.
 
+On FreeBSD you can `pkg install hs-elm-format` or use [devel/elm-format](https://www.freshports.org/devel/elm-format/), then execute `elm-format` normally.
```
avh4: Do you know why the package is called `hs-elm-format` instead of `elm-format` ?  I'd like to add some text about other package managers as well, and it would be great if there could be a consistent package name across systems.

evilham: I see your point, but it's a FreeBSD thing, basically everything haskell has a hs- prefix.
Same way every python or php package have a python or php prefix. I hope that's not an issue.
Btw 0.8.2 is already installable from FreeBSDs precompiled port packages.

avh4: ah, yeah, I think that's a bit confusing since elm-format is a command line tool that happens to be written in haskell and not a haskell library.  But thanks for the info.

evilham: I agree :-D one gets used to it in the OS though.
Thank you for merging and for elm-format!

evilham: I agree :-D one gets used to it in the OS though.
Thank you for merging and for elm-format!

evilham: I agree :-D one gets used to it in the OS though.
Thank you for merging and for elm-format!

PR #636, opened 2019-10-01
nixify upgrade tool development process

BrianHicks:
this makes the workflow mentioned in the README possible, except with Stack installed by nix.

I'm sure it'd be possible to make `shake` do the right thing too without `stack install`, but I didn't want to change the development flow any more than strictly necessary!

Note also that I'm not using the stack nix integration—you can try it with `stack --nix build` but it seems to slow things down for me in another project so I left it out here.


shell.nix
```diff
@@ -0,0 +1,12 @@
...
+let
+  sources = import ./nix/sources.nix;
+
+  niv = import sources.niv { };
+
+  nixpkgs = import sources.nixpkgs { };
+in with nixpkgs;
+stdenv.mkDerivation {
+  name = "elm-format";
+  buildInputs = [ niv.niv stack libiconv ];
```
avh4: why was libiconv needed?

BrianHicks:
I don’t recall but it wouldn’t build without it

avh4:
I notice this doesn't work with `nix-shell --pure`, mainly because gcc is missing, though I tried adding gcc and (git which is needed by the build script to determine the version number), and then `stack clean` and then trying to build everything in a pure shell went into some infinite loop complaining about a broken pipe when running the haskell tests.

with an approach like this, is it common to not try to include basic tools that would generally be installed outside of nix?

avh4:
> libiconv

oh, this must be a dependency of some of the haskell dependencies that have native code -- and it didn't break for me because my stack already has all those dependencies cached (outside of the project workspace)

BrianHicks:
Hmm, I couldn't replicate the missing `gcc` dependency. Thanks for the prompt to run with `--pure`, though, I found the shell was missing `git` and `python3`.

Here's the command I'm using to test. Does it match yours?

```
nix-shell --pure --run 'stack clean; stack runhaskell -- Shakefile.hs'
```

avh4:
Ah, maybe it was python I was missing...

avh4:
Did you include clearing `_build`?

Adding `gcc git python3` to the buildInputs, with `nix-shell --pure --run 'stack clean; rm -Rf _build; stack runhaskell -- Shakefile.hs'` I get:

```
elm-format> test (suite: elm-format-tests, args: --hide-successes)
                        
elm-format-tests: internal error: evacuate(static): strange closure type 258342912
    (GHC version 8.6.5 for x86_64_apple_darwin)
    Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug

elm-format> Test suite elm-format-tests failed
Completed 2 action(s).
Test suite failure for package elm-format-0.8.2
    elm-format-tests:  exited with: ExitFailure (-6)
Logs printed to console
```

avh4:
hmm, there's a lot of

```
ld: warning: object file (/Users/avh4/.stack/snapshots/x86_64-osx/.../8.6.5/lib/x86_64-osx-ghc-8.6.5/colour-2.3.5-.../libHScolour-2.3.5-....a(Matrix.o)) was built for newer OSX version (10.13) than being linked (10.12)
```

ah, I guess I need to wipe `~/.stack`, maybe

avh4:
With `nix-shell --pure --run 'export STACK_ROOT=/Users/avh4/workspace/elm-format3/.stack; stack setup --verbose; stack clean; rm -Rf _build; stack runhaskell -- Shakefile.hs'`, I get:

```
2019-10-04 11:20:49.177042: [error] /nix/store/qmmjygjchy30pkkcm56hiy6amp1nlafy-gnutar-1.32/bin/tar: Ignoring unknown extended header keyword 'LIBARCHIVE.creationtime'
Unpacking GHC into /Users/avh4/workspace/elm-format3/.stack/programs/x86_64-osx/ghc-                                                                                    2019-10-04 11:20:49.177399: [error] /nix/store/qmmjygjchy30pkkcm56hiy6amp1nlafy-gnutar-1.32/bin/tar: Ignoring unknown extended header keyword 'SCHILY.dev'
Unpacking GHC into /Users/avh4/workspace/elm-format3/.stack/programs/x86_64-osx/ghc-                                                                                    2019-10-04 11:20:49.177530: [error] /nix/store/qmmjygjchy30pkkcm56hiy6amp1nlafy-gnutar-1.32/bin/tar: Ignoring unknown extended header keyword 'SCHILY.ino'
Unpacking GHC into /Users/avh4/workspace/elm-format3/.stack/programs/x86_64-osx/ghc-                                                                                    2019-10-04 11:20:49.177644: [error] /nix/store/qmmjygjchy30pkkcm56hiy6amp1nlafy-gnutar-1.32/bin/tar: Ignoring unknown extended header keyword 'SCHILY.nlink'
Unpacking GHC into /Users/avh4/workspace/elm-format3/.stack/programs/x86_64-osx/ghc-

:20:52.087917: [debug] checking for path to top of build tree... /Users/avh4/workspace/elm-format3/.stack/programs/x86_64-osx/ghc-8.6.5.temp/ghc-8.6.5
2019-10-04 11:20:52.090287: [debug] checking for perl... no
2019-10-04 11:20:52.090465: [debug] You must install perl before you can continue
2019-10-04 11:20:52.090543: [debug] Perhaps it is already installed, but not in your PATH?
2019-10-04 11:20:52.139528: [error] Received ExitFailure 1 when running
Raw command: /Users/avh4/workspace/elm-format3/.stack/programs/x86_64-osx/ghc-8.6.5.temp/ghc-8.6.5/configure --prefix=/Users/avh4/workspace/elm-format3/.stack/programs/x86_64-osx/ghc-8.6.5/
Run from: /Users/avh4/workspace/elm-format3/.stack/programs/x86_64-osx/ghc-8.6.5.temp/ghc-8.6.5/

2019-10-04 11:20:52.139752: [error] 
Error: Error encountered while configuring GHC with
         /Users/avh4/workspace/elm-format3/.stack/programs/x86_64-osx/ghc-8.6.5.temp/ghc-8.6.5/configure --prefix=/Users/avh4/workspace/elm-format3/.stack/programs/x86_64-osx/ghc-8.6.5/
```

BrianHicks:
looks like it's building GHC… that shouldn't happen! Is stack doing that or nix?

avh4:
That's stack downloading (not building) ghc, because I cleared the STACK_ROOT folder

BrianHicks:
why is that only happening in a pure shell? I guess it could be removing HOME, but that level does not seem worth solving for.

avh4:
Oh, I see, you're right, it is trying to compile ghc -- I guess I don't understand why it's not downloading a prebuilt ghc.

Okay, so it seems like the required steps to use nix will be:
- install stack via some means
- run `stack setup`
- then use `nix-shell` as normal (which will get it's own version of stack, but that stack should reuse the already-installed ghc and snapshots that the non-nix `stack setup` installed into `~/.stack`

avh4:
Apparently adding `perl` to the buildInputs allos stack to successfully build ghc when it needs to!

BrianHicks:
Hmm, I think it’s building GHC because the pure shell cannot find the right HOME or something. Maybe that not working under pure is fine

avh4:
I added perl and comments here: https://github.com/avh4/elm-format/commit/53832730fe8095ef3b02dc533bd5346f09e02168

Thanks!

PR #648, opened 2019-11-16
Add support for single-line `if` expressions

### LINKED ISSUE #209: Should single-line `if` expressions be allowed?
Following code

``` elm
isPalindrome : List comparable -> Bool
isPalindrome list =
    List.map2 (,) list (List.reverse list)
        |> List.foldl
            (\( x, y ) acc -> if x == y then acc else False)
            True
```

turns to this

``` elm
isPalindrome : List comparable -> Bool
isPalindrome list =
    List.map2 (,) list (List.reverse list)
        |> List.foldl
            (\( x, y ) acc ->
                if x == y then
                    acc
                else
                    False
            )
            True
```

What do you think about it guys?

rlefevre:
Fixes #209


rlefevre:
Note: I don't know if there is a more optimized solution than the 3 `trackNewline`.

tests/test-files/transform/Elm-0.18/Examples.formatted.elm
```diff
@@ -2,31 +2,15 @@ module Main exposing (bar, multilineList, ratio)
...
-            True
-
-        else
-            False
-    then
-        "a"
-
-    else
-        "b"
+    if if a then True else False then "a" else "b"
```
rlefevre: Do we want this?

rl-king: I've never written code like that and have never seen any write code like that in reviews. Maybe because it's not very idiomatic functional code. I'm not against it, although I'd say it's a little less readable compared to the multiline version. 

If we want to discourage code like this by having it format in a certain way, I'm not sure if one or the other does a better job at that.

parser/src/Parse/Expression.hs
```diff
@@ -191,11 +191,18 @@ ifExpr elmVersion =
...
 
-      return $ E.If first rest final
+
+      return $ E.If first rest final $
+          case (firstMultiline, restMultiline, finalMultiline) of
+            (JoinAll, JoinAll, JoinAll) ->
+                multilineToBool JoinAll
+
+            _ ->
+                multilineToBool SplitAll
```
avh4: Ah, Bool was the old way of storing multiline... the new arg to `If` should be `Multiline` (from AST.V0_16) instead of `Bool`

tests/test-files/transform/Elm-0.18/Examples.formatted.elm
```diff
@@ -2,31 +2,15 @@ module Main exposing (bar, multilineList, ratio)
...
 
 ratio =
-    graphHeight
-        / (if range == 0 then
-            0.1
-
-           else
-            toFloat range
-          )
+    graphHeight / (if range == 0 then 0.1 else toFloat range)
```
avh4: Why did the examples in this file need to change?  Was it just to add test coverage of the new single-line case? Can these go back to how they were after the new cases are added to `Elm-0.19/AllSyntax/Expressions.elm`?

avh4: Oh I see, the input files for these tests were single-line.  That seems fine, then.

rlefevre:
I totally agree this needs more tests, but I need some guidance because I don't understand if tests in `ExpressionTest.hs` should use `example` that seem to only test parsing, `example'` that compares to the formatted string, or tests like `tests/test-files/transform/Elm-0.18/Examples.formatted.elm ` that compare a full file. I'm a little confused.

avh4:
Ah, `ExpressionTest.hs` is meant to be phased out (as you've probably noticed, it's a pain to update).  The `tests/test-files/good` is the main test suite, and every new feature should have tests there (for this change, I think `Elm-0.19/AllSyntax/Expressions.elm` makes sense.  Then `tests/test-files/transform` is for things that require the input file to be different from the output file.

So for this PR:
- add examples of the new allowed formats in `tests/test-files/good/Elm-0.19/AllSyntax/Expressions.elm`
- update `ExpressionTest.hs` as little as possible so that it compiles and passes
- If there's a `tests/test-files/transform` you want to add, you can, but it's not required

rlefevre:
One undiscussed point is does it make sense to allow single-line:
```
if {- A -} True then {- B -}  1 else {- C -} 2
```

For now, because it reuses some multi-line behavior for comments, my implementation keeps the following on a single line:
```
if {- A -} True then 1 else 2
```
But any comment in the `then` or `else` bodies would make it multi-line.
This does not seem logical and may not be the wanted design :thinking:

parser/src/Parse/Expression.hs
rlefevre: Changed to `Multiline`, I used a `Bool` at first because `Multiline` was not yet used in `parser/src/AST/Expression.hs`, which made me suspiscious of using it.

Note that `multilineToBool` in `parser/src/Parse/Helpers.hs` seems redundant with `isMultiline` from `parser/src/AST/V0_16.hs`, but both may be needed for some reason I did not understand.

tests/test-files/good/Elm-0.19/AllSyntax/Expressions.elm
```diff
@@ -273,6 +273,19 @@ lambdaWithMultilinePattern =
 ifStatement =
     let
         a =
+            if True then 1 else 2
+
+        b =
+            if {- B -} True then 1 else 2
```
rlefevre: As said in a comment, `if {- A -} True then {- B -} 1 else {- C -} 2` is not accepted on single-line, should it be?

avh4: Yeah, I think it makes sense for
```
if {- A -} True {- B -} then {- C -} 1 {- D -} else {- E -} 2
```
to stay on one line.  This can probably be done in your logic by using `formatCommented`, `formatPreCommented`, `formatPostCommented` to make a box that includes the comments first before destructuring to check for SingleLine.

tests/test-files/good/Elm-0.19/AllSyntax/Expressions.elm
```diff
@@ -273,6 +273,16 @@ lambdaWithMultilinePattern =
 ifStatement =
     let
         a =
+            if True then 1 else 2
```
avh4: For completeness, I think it would be good to add a single-line else if example, assuming that's allowed:

```
if True then 1 else if True then 2 else 3
```

avh4: Oh, nevermind, from the implementation code, that's not allowed to be single-line.

tests/test-files/good/Elm-0.19/AllSyntax/Expressions.elm
rlefevre: It was a little more complicated in order to avoid rendering twice the `then` and `else` bodies, I have therefore switched a top level `formatIfExpression` function.

Please tell me if you see some desirable formatting or code/splitting improvements.

tests/test-files/good/Elm-0.19/AllSyntax/Expressions.elm
rlefevre: Yeah, it seems to me that `else if` were not part of the use cases listed in the discussion and lack readability for single-line, I have therefore forbidden them.

That's also why I have added the `tests/test-files/transform/Elm-0.19/IfElseIfAlwaysMultiLine.elm` transform test.

tests/test-files/transform/Elm-0.18/Examples.formatted.elm
rlefevre: It think this is the last remaining point for now.
Should we change the example, change the implementation, or let it as is?

src/ElmFormat/Render/Box.hs
```diff
@@ -2120,6 +2072,106 @@ formatComment comment =
         AST.CommentTrickBlock c ->
             mustBreak $ row [ punc "{--", literal c, punc "-}" ]
 
+formatIfExpression ::
```
rlefevre: I don't know where I should put my function in the file.
It seemed to me at first that some main top level functions may have been sorted alphabetically, and that's why I put it between `formatComment*` and `formatLiteral`, but there are a lot of exceptions that could make this guess completely false.

tests/test-files/transform/Elm-0.18/Examples.formatted.elm
avh4: hmm, I guess it looks the original test was testing the indentation of the if statement in the predicate, so maybe the test input should change to have a single newline somewhere in the inner if statement.

rlefevre: The original commit of the test from 2015 is:
> Handle multiline if conditions

So it seems to have been made nested to stress the implementation.
Anyway, it seems better to keep it multiline, so I changed it to:
```
bar = (if if a then
  True else False then "a" else "b")
```

rlefevre:
I think that the PR is now complete, including tests.
I will just rebase it once the `upgrade-tool` branch is merged.

src/ElmFormat/Render/Box.hs
avh4: sounds fine.  I usually just put it directly below the function where I first use it 🤷‍♂

tests/test-files/transform/Elm-0.19/IfElseIfAlwaysMultiLine.elm
```diff
@@ -0,0 +1 @@
+x = if True then 1 else if False then 2 else 3
```
avh4: 👍

src/ElmFormat/Render/Box.hs
```diff
@@ -2120,6 +2072,104 @@ formatComment comment =
...
+        (AST.JoinAll, SingleLine cond', Right singleThens, [], Right singleElses)  ->
+            line $ row $ concat
+                [ [ keyword "if", space, cond', space, keyword "then", space ]
+                , List.intersperse space singleThens
+                , [ space, keyword "else", space ]
+                , List.intersperse space singleElses
+                ]
+
+        (_, ifCond, thenBody, _, elseBody) ->
+            formatIfExpressionMultiline elmVersion importInfo ifCond thenBody elseifs elseBody
```
avh4: The way these two functions work together seems a bit complicated.  The ideal is that these functions basically have two phases: 1) turning all the stuff into boxes and 2) assembling the boxes based on the multiline features.  Here we turn just enough into boxes to decide if we should call the other function, which makes it a bit harder to follow the full logic.

Though looking at the code for a few minutes, I don't have any quick suggestions of how to refactor this, so if you don't have time to give a try to simplifying this more, that's fine.  (A lot of the existing code in here is way more confusing than this)

avh4:
Looks great, thanks!  I'll let you know when I get `upgrade-tool` merged (goal is the next two weeks).

And also I think this will start the 0.9 releases as a notable behavior change.

src/ElmFormat/Render/Box.hs
rlefevre: I think it should be easier to follow the logic now.

src/ElmFormat/Render/Box.hs
```diff
@@ -1595,6 +1547,112 @@ formatExpression' elmVersion importInfo context aexpr =
...
+
+
+boxOrLineToBox :: Either Box Line -> Box
+boxOrLineToBox boxOrLine =
+    case boxOrLine of
+        Right singleLine -> line singleLine
+        Left box -> box
+
+
+boxesOrLinesToBox :: Bool -> Either [Box] [Line] -> Box
```
rlefevre: Maybe we could put this function and `boxOrLineToBox` to `Box.hs`?
I'm not sure because of the use of `ElmStructure.forceableSpaceSepOrStack1`.

andys8:
@rlefevre (and @avh4) Thanks so much for tackling this issue. It is really "improving the language" for me :)

ChristophP:
Really looking forward to this feature. I feel like this would clean up a bunch of code the previously required lots of space

Issue #647, opened 2019-11-16
Single/Multi-line formatting trigger discussion


rlefevre:
This is a discussion about the way we can trigger formatting on one line vs multi-line.

I'm sorry if this has been discussed before and I did not find it, if this is the case please just point me to the discussion.

I was thinking about https://github.com/avh4/elm-format/issues/507 and https://github.com/avh4/elm-format/issues/209 (which I'm all for because they are opt-in), and thought about the cost of converting some multi-line code to single-line code if they are implemented, particularly the `case...of` one.

Currently, for lists, tuples, records and lambdas, multi-line is triggered if the expression is on more than one line. This makes it very easy to get proper multi-line indentation from single-line with a single line break, but more complicated to get the reverse operation, particularly if the single line `case...of` patterns were implemented.

So I was thinking that using only the first indented element (symbol or expression) to trigger the wanted formatting would make both operations really easy, at a slight cost of feature discoverability. Basically, the main difference with current method is that positioning the first indented element would "suck" all the others in the multiline-line to single-line case.

# Examples
For example, by supposing that the PRs above are implemented:

### Lists
```
[1
, 2, 3, 4]
```
would be formatted as
```
[ 1
, 2
, 3
, 4
]
```        
### Tuples
```
(1
)
```
would be formatted as
```
( 1
)
```

### Records
```
{a=True
}
```
would be formatted as
```
{ a = True
}
```

### Case of
```
case x of
  1 ->
      foo1
  2 -> bar
```
would be formatted as
```
case x of
    1 ->
        foo1

    2 ->
        bar
```

### If then else
```
if cond then
t else f
```
would be formatted as
```
if cond then
    t

else
    f
```
### Lambda
```
\x ->
x + 1
```
would be formatted as
```
\x ->
  x + 1
```
Not much surprising as this is a subset of the current behavior.
The `,` or `]`, `)`, first case expression or `then` expression are on a different line, therefore the whole expression will be indented on multi-line.

## Leads to single-line
This becomes quite different from current behavior.
### Lists
```
[1,
2, 3, 4]
```
would be formatted as
```
[ 1, 2, 3, 4 ]
```

### Tuples
```
(1,
2, 3)
```
would be formatted as
```
( 1, 2, 3 )
```

### Records
```
{ a = True,
b = False}
```
would be formatted as
```
{ a = True, b = False }
```

### Case of
```
case x of
  1 -> foo1
  2 -> 
     bar2
  3 -> 
     bar3
```
[...]

PR #649, opened 2019-11-16
Allow single-line case-expressions branches

### LINKED ISSUE #507: "Single-line" case-expressions branches?
I noticed in the code for package.elm-lang.org that @evancz wrote [a case-expression](https://github.com/elm/package.elm-lang.org/blob/6d817cde29ddc94271d4b1b9614aa55895c5027c/src/frontend/Main.elm#L176-L180) where each branch is a single line:

```elm
case model.page of
    NotFound session -> session
    Search m -> m.session
    Docs m -> m.session
    Diff m -> m.session
```

`elm-format` currently formats this as:

```elm
case model.page of
    NotFound session ->
        session

    Search m ->
        m.session

    Docs m ->
        m.session

    Diff m ->
        m.session
```

Is this an improvement? I found the former easier to understand because I could spot the pattern more easily. At a glance I could more quickly tell "ah, it's extracting the `session` in each branch."

There are many situations in which `elm-format` gives the author a binary choice: single-line mode or multiline mode. Should there be a similar choice for case-expressions? That is, allow authors to choose between a case-expression with all single-line branches (e.g. for "mappings" like the above), versus a case-expression with all multiline branches like we have today?

rlefevre:
This implementation uses only the first case branch to decide if single-line branches is wanted.
This allows to easily go from multi-line to single-line, even with lots of branches.

Closes #507


rlefevre:
Note: this implementation uses the method proposed in https://github.com/avh4/elm-format/issues/647 to trigger the single-line branches.

src/ElmFormat/Render/Box.hs
```diff
@@ -1493,51 +1494,57 @@ formatExpression' elmVersion importInfo context aexpr =
...
                           stack1
                               [ line $ keyword "case"
                               , indent subject'
                               , line $ keyword "of"
                               ]
 
-                clause (pat, expr) =
+                clausesBodies =
+                    map (formatHeadCommentedStack (formatExpression elmVersion importInfo SyntaxSeparated) . snd) clauses
+
```
rlefevre: Maybe I should use `mapAccumR` instead of `map` in `clause` and `foldl'` in `splitAllClauses`.

rlefevre:
Updated with `Multiline` instead of `Bool`, and tests added.

rlefevre:
@avh4 Given what's accepted with single-line `if` expressions, do we want to accept single-line comments in branches like the following example?
```
case Just 1 of
    Just x -> {- A -} x
    _ -> {- B -} 2
```
For now this is not supported in this PR (comments lead to multi-line except those between `case...of`).

avh4:
> do we want to accept single-line comments

yes, I think so

rlefevre:
Ok. In multi-line, I noticed that for if expression we get:
```
            if {- C -} True {- D -} then      
                {- E -} {- F -}      
                1      
```
so several comments are on the same line, whereas for case branches, we get:
```
                ->                                                                         
                    {- Q -}
                    {- R -}
                    x

```
with each comment put it own line.

Is it on purpose and why?

src/ElmFormat/Render/Box.hs
```diff
@@ -1493,65 +1493,9 @@ formatExpression' elmVersion importInfo context aexpr =
...
-                              ]
-
-                clause (pat, expr) =
-                    case
-                      ( pat
-                      , (formatPattern elmVersion False $ (\(AST.Commented _ x _) -> x) pat)
-                          |> negativeCasePatternWorkaround pat
-                      , formatCommentedStack (formatPattern elmVersion False) pat
-                          |> negativeCasePatternWorkaround pat
-                      , formatHeadCommentedStack (formatExpression elmVersion importInfo SyntaxSeparated) expr
```
rlefevre: Performance should be slightly better because we don't format twice the pattern anymore.

tests/test-files/good/Elm-0.19/AllSyntax/Expressions.elm
```diff
@@ -362,24 +377,68 @@ caseStatement =
...
+                --F
+                _
+                --G
+                ->
+                    --H
+                    2
+
+        g =
+            case {- A -} {- B -} Just 1 {- C -} {- D -} of
+                {- E -} {- F -}
```
rlefevre: This seems more consistent with `if` expressions.

rlefevre:
I think that the PR is now complete, including tests.
I will just rebase it once the `upgrade-tool` branch is merged.

avh4:
> so several comments are on the same line, whereas for case branches, we get

I suspect there was not a reason for this difference.

```
case Just 1 of
    {- A -} Just x {- B -} -> {- C -} x
```

comments in position A are very rare from code I've observed, so I don't think it really matters how we choose to deal with these.

parser/src/Parse/Expression.hs
```diff
@@ -234,8 +234,8 @@ caseExpr elmVersion =
       (e, multilineSubject) <- trackNewline $ (\(pre, e, post) -> Commented pre e post) <$> padded (expr elmVersion)
       reserved elmVersion "of"
       firstPatternComments <- whitespace
-      result <- cases firstPatternComments
-      return $ E.Case (e, multilineToBool multilineSubject) result
```
avh4: 👍 thanks for upgrading the existing Bool

parser/src/Parse/Expression.hs
```diff
@@ -234,8 +234,8 @@ caseExpr elmVersion =
...
+      (branches, multilineFirstBranch) <- cases firstPatternComments
+      return $ E.Case (e, multilineSubject) (branches, multilineFirstBranch)
```
avh4: Was it in a separate issue you were suggesting using the linebreak of the first item to control multiline for other kinds of expressions besides cases?  I'm still hesitant about that.  I'd be more comfortable having this only be `JoinAll` if *all* of the branches are single line,  and then maybe doing a separate experiment after adding this to try the idea of having the first item control everything.

src/ElmFormat/Render/Box.hs
```diff
@@ -2121,6 +2065,87 @@ formatComment comment =
...
+              (_, subject') ->
+                  stack1
+                      [ line $ keyword "case"
+                      , indent subject'
+                      , line $ keyword "of"
+                      ]
+
+        (multilineBranches, branches') =
+            List.mapAccumR
+            (\mline (AST.Commented prePat pat postPat, (preRes, res)) ->
```
avh4: not necessary to change before merging, but FYI my preferred style here is to take this anonymous function and define it in the outermost let block right above the `(multilineBranches, branches') =` definition.

src/ElmFormat/Render/Box.hs
```diff
@@ -2121,6 +2065,87 @@ formatComment comment =
...
+                        Maybe.maybeToList $ formatComments prePat
+                    pat' =
+                        formatPattern elmVersion False pat
+                            |> negativeCasePatternWorkaround pat
+                    postPat' =
+                        Maybe.maybeToList $ formatComments postPat
+                    preRes' =
+                        Maybe.maybeToList $ formatComments preRes
+                    res' =
+                        formatExpression elmVersion importInfo SyntaxSeparated res
```
avh4: `res`/`res'`/`preRes` should I think be called "body" instead of "res" to be consistent with other code, unless there was a specific reason you chose that name.

src/ElmFormat/Render/Box.hs
```diff
@@ -2121,6 +2065,87 @@ formatComment comment =
...
+                    stack1
+                      [ stack1 $ prePat ++ [pat] ++ postPat
+                      , line $ keyword "->"
+                      , indent $ stack1 $ preRes ++ [res]
+                      ]
+    in
+    opening
+        |> andThen
+            (branches'
+                |> map (branch multilineBranches)
```
avh4: Not necessary to change before merging, but FYI I'm trying to prefer `fmap` over `map`

src/ElmFormat/Render/Box.hs
```diff
@@ -2121,6 +2065,87 @@ formatComment comment =
...
+                |> (if AST.isMultiline multilineBranches then List.intersperse blankLine else id)
+                |> map indent
```
avh4: Will this indent the blank lines?  I can't remember if that matters...  if this ends up leaving empty lines containing spaces, then that needs to be avoided

src/ElmFormat/Render/Box.hs
```diff
@@ -2121,6 +2065,87 @@ formatComment comment =
...
+                        (mline, (singles, prePat', pat', postPat', preRes', res'))
+                    _ ->
+                        (AST.SplitAll, (singles, prePat', pat', postPat', preRes', res'))
+            )
+            multilineFirstBranch
+            branches
+
+        branch multiline' (singles, prePat, pat, postPat, preRes, res) =
+            case (multiline', singles, prePat, pat, postPat) of
+                (AST.JoinAll, Right singleLines, _, _, _) ->
```
avh4: I see my previous code was similarly confusing, but it's not obvious here that `singleLines` includes the data from `pat, postPat, preRes, res`.  Could you at least add a comment right above this `case` noting that?  (without it, I started reading these branches and though "oh no, is some of the AST getting lost here?")

src/ElmFormat/Render/Box.hs
rlefevre: Good point. I will check.

parser/src/Parse/Expression.hs
rlefevre: I thought we could experiment it with only `case` expressions, but you're right, it's better to introduce the single `if` and `case` expressions in a consistent way with existing behavior, then discuss eventually the other method. I will make the change.

rlefevre: New version uses JoinAll if all branches are single line.

src/ElmFormat/Render/Box.hs
rlefevre: fixed

src/ElmFormat/Render/Box.hs
rlefevre: Fixed

src/ElmFormat/Render/Box.hs
rlefevre: Fixed

src/ElmFormat/Render/Box.hs
rlefevre: I tested and blank lines are trimmed to only "\n". 
I have not checked where though, most likely during `Box -> Text` rendering.

src/ElmFormat/Render/Box.hs
rlefevre: I have reworked this part and it is hopefully a lot less confusing.
I used an `Either` to differentiate between single lines and boxes to avoid the variables overlap.

tommyengstrom:
Nice! I was just about to stop using elm-format after having used it for some time. I find that all the extra lines makes the code much harder to read than if I wasn't using anything at all. This fixes the main problem. :+1:

Issue #339, opened 2017-03-29
Make sure the normalized markdown is compatible with elm-markdown


avh4:
http://package.elm-lang.org/packages/evancz/elm-markdown/latest is what's used on package.elm-lang.org.  So the normalized markdown needs to cause no ambiguities with that engine.

Comparison tool: https://johnmacfarlane.net/babelmark2/

Possibly use the marked js library in integration tests to verify things.

avh4:
The fix I want for this is a test suite that does this validation.  I'm going to remove this from the 0.8.0 release, as there haven't yet been any bug reports about any issues.

gampleman:
One issue I found is that elm-format transoforms:

```markdown
[Scale](Scale)
```

into 

```markdown
<Scale>
```

which elm-markdown renders as 

&lt;Scale&gt;

rather than

[Scale](Scale)

PR #651, opened 2019-11-17
Never transform links to autolinks

rlefevre:
This fixes one bug raised in https://github.com/avh4/elm-format/issues/339
that was causing links to modules like `[Module](Module)` being
transformed to invalid autolinks like `<Module>`.

### DIFF
--- a/markdown/Cheapskate/Inlines.hs
+++ b/markdown/Cheapskate/Inlines.hs
@@ -415,15 +415,14 @@ pAutolink = do
          | otherwise   -> fail "Unknown contents of <>"
 
 autoLink :: Text -> Inlines
-autoLink t = singleton $ Link (toInlines t) (Url t) (T.empty)
-  where toInlines t' = case parse pToInlines t' of
-                         Right r   -> r
-                         Left e    -> error $ "autolink: " ++ show e
-        pToInlines = mconcat <$> many strOrEntity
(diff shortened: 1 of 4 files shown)


markdown/Cheapskate/Inlines.hs
```diff
@@ -415,15 +415,14 @@ pAutolink = do
...
 autoLink :: Text -> Inlines
-autoLink t = singleton $ Link (toInlines t) (Url t) (T.empty)
-  where toInlines t' = case parse pToInlines t' of
-                         Right r   -> r
-                         Left e    -> error $ "autolink: " ++ show e
-        pToInlines = mconcat <$> many strOrEntity
+autoLink t = 
+  case parse pToInlines t of
+      Right _   -> singleton $ Autolink t
+      Left e    -> error $ "autolink: " ++ show e
```
rlefevre: To be honest, I fail to think of a case where this would fail thinking
So I don't full understand this part of the code.

rlefevre:
I have added a `tests/test-files/good/Elm-0.19/AllSyntax/Doc.elm ` test file, but I'm not sure how to organize it. If you have a better idea tell me.

avh4:
0.8.4 is now released with this fix: https://github.com/avh4/elm-format/releases/tag/0.8.4

PR #653, opened 2019-11-25
Autodetect elm version per file

### LINKED ISSUE #561: auto-detect elm version per file, instead of based on cwd
In 0.8.1, the elm-version is auto-detected based on the presence of `elm.json` or `elm-package.json` in the current working directory.

Proposal: instead of checking the current working directory, for each file check the file's folder and if not found traverse upwards until a package file is found.

It seems like this would be more flexible, and still do the right thing in all current cases?

rlefevre:
Resolves #561


rlefevre:
To avoid unexpected behavior changes:

* Per file detection is used only:
   - when elm-format does not run with `--upgrade`
   - there is no `--elm-version` flag
* Auto-detection from the directory where elm-format is run is still used to determine `--upgrade` versions and with `--stdin` if no `--elm-version` flag is provided.

src/ElmFormat.hs
```diff
@@ -83,23 +86,26 @@ resolveFiles inputFiles =
                 return $ Left ls
 
             Right files ->
-                return $ Right $ concat files
+                return $ Right $ concat $ files
+
 
+type ElmFile
```
rlefevre: Would you prefer a `newtype` instead? Or no type at all and just using the tuple everywhere?

src/ElmFormat.hs
```diff
@@ -234,44 +241,24 @@ main'' elmFormatVersion_ experimental_ args =
                             exitWithError message
 
                         Right whatToDo -> do
-                            elmVersionChoice <- case Flags._elmVersion config of
```
rlefevre: There was no point to use an `Either` anymore, as only `Right` were used.

src/ElmFormat/Filesystem.hs
```diff
@@ -63,3 +64,7 @@ findAllElmFiles inputFile =
 hasFilename :: String -> FilePath -> Bool
 hasFilename name path =
     name == FilePath.takeFileName path
+
+addElmVersion :: FileStore f => FilePath -> Free f (FilePath, ElmVersion)
```
rlefevre: You might prefer other names for this function and `ElmVersion.fromFile`, I'm open to suggestions if you don't like them.

src/ElmVersion.hs
```diff
@@ -58,3 +61,22 @@ style_0_19_cannotExposeOpenListing elmVersion =
         Elm_0_18 -> False
         Elm_0_18_Upgrade -> False
         _ -> True
+
+fromFile :: FileStore f => FilePath -> Free f ElmVersion
+fromFile path =
+    do
+        let dir = takeDirectory (path)
+        elmPackageJson <- stat (path </> "elm-package.json")
```
rlefevre: Using `path` instead of `dir` allows to also work with directories like `"."`, the small drawback is an additional test for files. We could test if it's a directory instead, but this makes the logic more complicated. As you wish.

rlefevre: Also `doesDirectoryExist` does not exist in `TestWorld.hs`, so this would require a little more work.

rlefevre:
The following instructions from `README.md` should also be updated:
> You must run elm-format from the directory that contains your elm.json (for Elm 0.19) or elm-package.json (for Elm 0.18), or else you must pass the appropriate --elm-version=0.19/--elm-version=0.18 command line argument.

Maybe you have a suggestion or this can be done later when preparing the release?

avh4:
> The following instructions from README.md should also be updated

I guess if we're not doing that now (because we want `master`'s README to still describe the existing version, could you maybe make a new github issue to track that, and I'll put it at the end of the next milestone

src/ElmFormat.hs
avh4: Yeah, I think either `newtype` or `data` with named record fields is preferable.

src/ElmFormat.hs
```diff
@@ -223,6 +229,7 @@ main'' elmFormatVersion_ experimental_ args =
                 do
                     let autoYes = Flags._yes config
                     resolvedInputFiles <- Execute.run (Execute.forHuman autoYes) $ resolveFiles (Flags._input config)
+                    detectedElmVersion <- Execute.run (Execute.forHuman autoYes) $ ElmVersion.fromFile "."
```
avh4: Can this IO action fail?  If so, we should maybe postpone running this until we know that `detectedElmVersion` will be used.

src/ElmFormat.hs
```diff
@@ -363,33 +350,41 @@ logErrorOr fn result =
             fn value *> return True
 
 
-doIt :: (InputConsole f, OutputConsole f, InfoFormatter f, FileStore f, FileWriter f) => ElmVersion -> WhatToDo -> Free f Bool
-doIt elmVersion whatToDo =
+
+doIt :: (InputConsole f, OutputConsole f, InfoFormatter f, FileStore f, FileWriter f) => ElmVersion -> Flags.Config -> WhatToDo -> Free f Bool
+doIt elmVersion config whatToDo =
+    let
+        getVersion fileDetectedElmVersion =
```
avh4: This feels like it's starting to get messy to me.  I'm thinking that we should probably try removing the `elmVersion` param to `doIt` and pushing it into each of the `WhatToDo` constructors that need it, and then we should also resolve the actual versions for each file (this `getVersion` logic) before calling `doIt`.  What do you think?

src/ElmFormat/Filesystem.hs
avh4: Yeah, the name seems good.  The one thing I wonder about is how it might work out to have `findAllElmFiles` return `[ElmFile]` instead of `[FilePath]` and having to use `addElmVersion` later.

avh4: Also having `findAllElmFiles` return `[ElmFile]` would allow a future optimization of detecting the version at the top and know it already as you recurse down into directories instead of later having to recurse up from each file to find its version.

src/ElmVersion.hs
```diff
@@ -58,3 +61,22 @@ style_0_19_cannotExposeOpenListing elmVersion =
...
+        case elmPackageJson of
+            IsFile ->
+                do
+                    elmJson <- stat (path </> "elm.json")
+                    return $ case elmJson of
+                        IsFile -> Elm_0_19
+                        _ -> Elm_0_18
+
+            _ | path == dir ->
+                return Elm_0_19
```
avh4: I think this branch means that if there is not `elm-package.json`, and we've recursed out as far as we can, then give Elm_0_19.  Is that right?   I think it should probably return `Nothing` in this case and let the caller decide what to default to rather than hardcoding the default in this function -- is there anything I'm missing about this?

avh4: Oh, was this done to only have to check one file per level as you move outward?

src/ElmVersion.hs
```diff
@@ -58,3 +61,22 @@ style_0_19_cannotExposeOpenListing elmVersion =
         Elm_0_18 -> False
         Elm_0_18_Upgrade -> False
         _ -> True
+
+fromFile :: FileStore f => FilePath -> Free f ElmVersion
+fromFile path =
+    do
+        let dir = takeDirectory (path)
+        elmPackageJson <- stat (path </> "elm-package.json")
+        case elmPackageJson of
```
avh4: Since we are now recursing upward (and it looks like we normalize the path, so we'll recurse up all the way to the root of the filesystem?) so that means we might go up to a folder that contains an `elm.json` or `elm-package.json` that is owned by a different user and maybe not readable by the current user.  Will `stat` still succeed in that case?  Are there any possible permissions that might cause an error? (Maybe we reach a folder that we can't list? though I'm not sure that would be possible given we are in a folder that's inside it.)  If there are possible IO errors from the `stat` due to files that would be outside of the current user's control, I think we ought to make sure we ignore them rather than crashing.

tests/Integration/CliTest.hs
```diff
@@ -46,6 +46,13 @@ tests =
                 |> TestWorld.uploadFile "elm-package.json" "{\"elm-version\": \"0.18.0 <= v < 0.19.0\"}"
                 |> run "elm-format" ["test.elm", "--validate"]
                 |> expectExit 0
+            , testCase "for mixed Elm 0.18 and 0.19" $ world
+                |> TestWorld.uploadFile "0.18/src/test.elm" "module Main exposing (f)\n\n\nf =\n    '\\x2000'\n"
+                |> TestWorld.uploadFile "0.18/elm-package.json" "{\"elm-version\": \"0.18.0 <= v < 0.19.0\"}"
+                |> TestWorld.uploadFile "0.19/src/test.elm" "module Main exposing (f)\n\n\nf =\n    '\\u{2000}'\n"
+                |> TestWorld.uploadFile "0.19/elm.json" "{\"elm-version\": \"0.19.0 <= v < 0.20.0\"}"
+                |> run "elm-format" ["0.18/src/test.elm", "0.19/src/test.elm", "--validate"]
+                |> expectExit 0
```
avh4: 👍

avh4:
Looks like a great start, thanks!

src/ElmVersion.hs
rlefevre: Ok. This first naive implementation has at least the merits of raising a few potential design issues, so let's slow down and go back to agreeing on what we want.

First, there does not seem to be an universal agreement of what `normalize` means for paths, but [`System.FilePath.normalise`](https://hackage.haskell.org/package/filepath-1.4.2.1/docs/System-FilePath-Posix.html#v:normalise)  used in the PR does not make a path an absolute one. This can be noticed by the fact that it does not use some IO ([`Sytem.Path.toAbsoluteFilePath`](http://hackage.haskell.org/package/paths-0.2.0.0/docs/System-Path.html#v:toAbsoluteFilePath) does). [`System.FilePath.takeDirectory`](http://hackage.haskell.org/package/filepath-1.4.2.1/docs/System-FilePath-Posix.html#v:takeDirectory) does not make the path absolute either.

> and it looks like we normalize the path, so we'll recurse up all the way to the root of the filesystem

We don't make the path an absolute path (the only use of `normalise` is in `TestWorld.hs` and this does not make a path absolute as explained previously), so we won't recurse all the way except if the path is already an absolute path. 
`takeDirectory` is only a string manipulation function, it has not real knowledge of the filesystem.
Consequently, for example:
* `./src/file.elm` will stop at `.`
* `/tmp/file/elm` will stop at `/`
* `../test.elm` will stop at `..`
* `..` will actually only test `.` because surprisingly `takeDirectory ".." == "."` :thinking: 

Consequently, if `elm-format `is run from a directory below the `elm.json` or `elm-package.json` file, the version may not be correctly detected. 

For example if in a 0.18 package, `elm-format .` is run from the `src` sub-directory (instead of the package root directory with the `elm-package.json` inside), this will return the default value 0.19.

After your review, it seems to me that preferably, the design should be instead:
* For each file argument:
  - make it absolute (which is actually also an operation that recursively walks upward up to `/`)
  - search for `elm-package.json` or `elm.json` upward up to `/`, trying to detect an upward version
  - then if it is a directory, continue recursively downward to search `elm` files, detecting `elm-package.json` and `elm.json` on the way, and therefore downward versions (in an optimized way), once per directory.

From a performance point of view, this would optimize the `findAllElmFiles` case (a directory argument), but would not optimize the detection of several files arguments, but I think the main use cases are a single file argument or a directory, so this could be acceptable?

A significant downside of this approach is that this would require to really simulate directories in `TestWorld.hs`, which seems quite daunting (also what is the directory above `.` in `TestWorld`?), or to hack a very small working use case for what already exists and do real tests instead for more complicated cases. None seems completely compelling but a compromise would have to be found. Current implementation is simple in `TestWorld` because we don't use absolute paths and only use string operations on paths, which did not require to add filesystem operations in `FileStore.hs`. Maybe I am just lazy :sweat_smile:.

@avh4, what do you think?

avh4:
Ah, well my comments were based on the assumption that `normalise` was getting to an absolute path.  I'm not actually sure what's necessary...

There are three main uses of elm-format: use from editors and editor plugins, use from the command line, and use from CI.  For editors, is it common for editors to be running elm-format from a working directory that's inside of the folder containing the elm.json?  I think probably not, though maybe I'm wrong about that.  For command line use, I can imagine someone might `cd` into a subfolder and want to run elm-format just on files under that folder.  I think that would require going up through the actual filesystem and not just removing segments from the path.  Is that a case we want to support?  As for CI use, I don't think elm-format would typically be run from the project folder or from the folder where the elm.json is.

src/ElmFormat.hs
rlefevre: Ok, I used a data with named record fields.

src/ElmFormat/Filesystem.hs
rlefevre: I added the optimization in the last version.

src/ElmFormat.hs
rlefevre: Ok, done.
Things could be even simpler if we get rid of global auto-detection and just use 0.19.
To be discussed, see my other comment about that.

src/ElmFormat.hs
rlefevre: This is still needed in last version, but things have changed a lot, so let's discuss it again based on the new version.

src/ElmVersion.hs
rlefevre: Fixed in last version.

rlefevre:
I made a new version that assumes that running elm-format from command line in a sub-directory of the Elm project should work. Consequently, elm version detection is done first upward, then downward during files finding.

For now I have kept the global current directory Elm version detection to avoid behavior regressions, but I'm really not sure it makes sense anymore. It seems that the only use case is running in an 0.18 project with `--stdin` and without `--elm-version`.

src/ElmFormat.hs
```diff
@@ -222,9 +228,11 @@ main'' elmFormatVersion_ experimental_ args =
             Just config ->
                 do
                     let autoYes = Flags._yes config
-                    resolvedInputFiles <- Execute.run (Execute.forHuman autoYes) $ resolveFiles (Flags._input config)
+                    currentDirectoryElmVersion <- Execute.run (Execute.forHuman autoYes) $ FS.findElmVersion "."
```
rlefevre: I'm not sure this is useful anymore versus just using 0.19 as the default version.

I kept current working directory global version detection to avoid behavior regressions, but I fail to see valid use cases for it. The only one I see is running in a 0.18 project with `--stdin` and without `--elm-version`, is it worth it?

src/ElmFormat/Filesystem.hs
```diff
@@ -2,10 +2,20 @@ module ElmFormat.Filesystem where
...
+
+import qualified Data.Maybe as Maybe
 import qualified System.FilePath as FilePath
 
 
+data ElmFile
+    = ElmFile
+        { version :: ElmVersion
+        , path :: FilePath
+        } deriving (Show)
```
rlefevre: Just used `deriving Show` for my own private tests. It can be removed if you want.

src/ElmFormat/World.hs
```diff
@@ -44,9 +47,16 @@ instance World IO where
     writeFile = Prelude.writeFile
     writeUtf8File path content = ByteString.writeFile path $ Text.encodeUtf8 content
 
-    doesFileExist = Dir.doesFileExist
-    doesDirectoryExist = Dir.doesDirectoryExist
+    doesFileExist path = do
+        exists <- try (Dir.doesFileExist path) :: IO (Either IOException Bool)
```
rlefevre: This is to avoid exceptions if detecting the version upward fails (because of permissions issues or something).

Maybe you would have preferred using `try` inside the Free Monad?
If so I would need guidance as I'm not sure how to do this with `Free` monad.

src/Messages/Formatter/HumanReadable.hs
```diff
@@ -41,8 +41,8 @@ renderInfo :: World m => InfoMessage -> m ()
 renderInfo (ProcessingFile file) =
     putStrLn $ "Processing file " ++ file
 
-renderInfo (FileWouldChange file) =
-    putStrLn $ "File would be changed " ++ file
+renderInfo (FileWouldChange file version) =
+    putStrLn $ "File would be changed " ++ file ++ " (" ++ show version ++ ")"
```
rlefevre: Is it ok to you? It seems useful to add the detected version in output.

tests/ElmFormat/TestWorld.hs
```diff
@@ -43,20 +44,28 @@ fullStderr state =
...
         do
             state <- State.get
-            return $ Dict.member path (filesystem state)
+            return $ Dict.member (FilePath.normalise path) (filesystem state)
 
     doesDirectoryExist _path =
         return False
 
+    makeAbsolute path =
+        -- wrong but enough for tests
```
rlefevre: Is it ok to you? I'm not sure that writing a lot of code for this now is worth it.

src/ElmFormat.hs
```diff
@@ -133,48 +155,42 @@ determineDestination output doValidate json =
...
         ( FromFiles _ _, ToJson ) -> Left SingleOutputWithMultipleInputs
 
 
-determineWhatToDoFromConfig :: Flags.Config -> Either [InputFileMessage] [FilePath] -> Either ErrorMessage WhatToDo
-determineWhatToDoFromConfig config resolvedInputFiles =
+determineWhatToDoFromConfig :: Flags.Config -> ElmVersion -> Either [InputFileMessage] [ElmFile] -> Either ErrorMessage WhatToDo
+determineWhatToDoFromConfig config defaultElmVersion resolvedInputFiles =
     do
-        source <- determineSource (Flags._stdin config) resolvedInputFiles
+        checkUpgradeVersion (Flags._upgrade config) (Flags._elmVersion config)
```
rlefevre: In 0.8.2, upgrade can actually use the auto-detected current directory elm version, which seems confusing to me (ie: `--upgrade ` is possible without `--elm-version`):

`MustSpecifyVersionWithUpgrade` says:
> "I can only upgrade code to specific Elm versions.  To make sure I'm doing what you expect, you must also specify --elm-version=" ++ show elmVersion ++ " when you use --upgrade."


This check makes sure that a flag is provided, whatever the result of "current directory elm version" (which might be removed) or "per file version detection" (which is overridden by the flag).

I'm not sure if this is the expected behavior.

src/ElmFormat.hs
```diff
@@ -109,16 +114,33 @@ data Destination
...
         ( True, Right (_:_) ) -> Left TooManyInputs
 
 
+upgradeVersion :: Bool -> ElmVersion -> ElmVersion
+upgradeVersion upgrade version =
+    case (upgrade, version) of
+        (True, Elm_0_18) ->
+            Elm_0_18_Upgrade
+
+        (True, _) ->
```
rlefevre: `checkUpgradeVersion` makes actually impossible to use something else than 0.19.

Still, the whole upgrade version code (check and upgrade) seems a little messy, but I'm not sure of the exact behavior we want, nor how to refactor this.

src/ElmFormat/World.hs
```diff
@@ -44,9 +47,16 @@ instance World IO where
...
+        return $ Either.fromRight False exists
+
+    doesDirectoryExist path = do
+        exists <- try (Dir.doesDirectoryExist path) :: IO (Either IOException Bool)
+        return $ Either.fromRight False exists
+
     listDirectory = Dir.listDirectory
+    makeAbsolute = Dir.makeAbsolute
```
rlefevre: This might also fail but I'm not sure what to do or if we want to handle this :thinking:

PR #667, opened 2020-02-19
Refer to elm-mode for instructions on customization

woochica:
Enabling elm-format in Emacs via [`M-x customize-variable elm-format-on-save` is obsolete](https://github.com/jcollard/elm-mode/blob/master/elm-format.el#L31).


avh4:
Thanks!  I wanted to keep the actual step here in the README rather than making people follow through multiple pages to get things set up, but I update the instructions for the new config here: https://github.com/avh4/elm-format/pull/701/commits/5f81628c856fae039e22eb2c476a8cd5c60f3398  I hope that's better!

PR #672, opened 2020-04-08
Use 2 spaces instead of 4

Unisay:
As agreed here https://github.com/tricycle/bellroy/pull/3030


samuelgiles:
👀

harrysarson:
Making pull requests against the wrong version of a repository is the bane of my life too

PR #691, opened 2020-09-01
in doc comments, inline code containing backticks is now escaped corr…

avh4:
…ectly


avh4:
CI ❌ is because Windows CI is broken right now.

PR #740, opened 2021-07-06
WIP: Parsec API for the new Elm parser

emmabastas:
This is PR is the first step in an effort to integrate the parser from the Elm compiler into elm-format as part of a GSOC project ([project proposal](https://github.com/elm-tooling/gsoc-projects/issues/13)).

This PR removes the dependency on [parsec](https://hackage.haskell.org/package/parsec) and [indents](https://hackage.haskell.org/package/indents-0.3.3) which have been used by elm-format to parse Elm code. The API these libraries provided has been replaced with drop-in replacement modules (located in `Text.Parsec.*`) compatible with elm-format. These replacement modules are implemented in terms of the new Elm parser (located in `Parse.Primitives` and copied from the Elm compiler). `new-parser-2021-notes.md` in the project root goes into more detail on some specifics relating to this project.

~~The entire test suite passes.~~

### DIFF
--- a/elm-format-lib/src/Text/Parsec/Char.hs
+++ b/elm-format-lib/src/Text/Parsec/Char.hs
@@ -16,7 +16,16 @@ module Text.Parsec.Char
   , string
   ) where
 
-import Text.Parsec.Prim (Parser)
+import qualified Parse.Primitives as EP
+import Text.Parsec.Prim (Parser(..))
+import Text.Parsec.Pos (newPos)
+import Text.Parsec.Error (Message(SysUnExpect, Expect), newErrorMessage, setErrorMessage)
+
+import Foreign.Ptr (plusPtr)
+import Data.Bits ((.|.), (.&.), shiftL, shiftR)
+import Data.Word (Word8)
+import Data.Char (chr, ord)
+
 
 oneOf :: [Char] -> Parser Char
 oneOf = undefined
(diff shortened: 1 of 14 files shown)


emmabastas:
Turns out not all tasts are passing after all.
All of the units test are passing, but only some integrations tests. In particular most code containing `let .. in` and `case .. of` fails to parse.

elm-format-lib/src/Text/Parsec/Char.hs
```diff
@@ -54,5 +63,120 @@ anyChar             = satisfy (const True)
...
+        0x0a {- "\n" -} -> (row + 1, 1)
+
+        -- The doccumentation for `Text.Parsec.Char.updatePosChar` claims that
+        -- carrige return ("\r") increments row by 1, just like newline ("\n").
+        -- This doesn't appear to be the case from looking at the code:
+        -- https://hackage.haskell.org/package/parsec-3.1.14.0/docs/src/Text.Parsec.Pos.html#updatePosChar
+        --
+        -- Let's not devle into this unless it turns out that elm-format
+        -- needs it.
+        0x0d {- "\r" -} -> error "Can't handle carrige return"
```
avh4: If I remember correctly, elm-format will never output CRs, but it does read them.  For ones in literal strings, it will convert them to an escape sequence.  And for ones elsewhere, it will treat them either as newlines or whitespace (I forget which).

PR #746, opened 2021-08-03
Use code from the Elm compiler to parse strings

emmabastas:
This PR focuses on migrating string and char literal parsing to use code from the Elm compiler.

7781377 Starts by copying over code from the Elm compiler verbatim. Some declarations that won't be needed are commented out, and `Reporting.Error.Syntax` is merged with the existing code in that module.

The major changes take place in 67d646d and 1a21c57 where the behaviors of `Elm.String` and `Parse.String`  are changed. The `ES.String`'s returned by `Parse.String.string` and `Parse.String.character` no longer represents raw strings as they should be written to JS, instead they represent the input string, but _normalized_ so that escape sequences have been replaced by the actual Unicode characters they represent. Input: `🐂\u{1F402}\x1F402`. Returned by parser: `f0 9f 90 82 f0 9f 90 82 f0 9f 90 82`

`Parse.Literal`'s `chr` and `str` are changed to use the new parsers in c2e023f (and again in 1a21c57)  and 4fcffba respectively.

f20232d, db8e002 and ~a92f6a8~ 356ce1b just reintroduces parsing features specific to elm-format


elm-format-lib/src/Parse/Literal.hs
```diff
@@ -126,7 +126,8 @@ chr =
     toError e = newErrorUnknown ("Error parsing char: " ++ show e)
   in
   do  s <- character toExpecation toError
-      processAs charLiteral $ sandwich '\'' (ES.toChars s)
+      case ES.toChars s of
+        [ c ] -> return c
```
avh4, line 102: We should add a "please report ..." error for the fallback branch.

elm-format-lib/src/Parse/String.hs
```diff
@@ -265,31 +257,30 @@ multiString pos end row col initialPos sr sc revChunks =
 -- ESCAPE CHARACTERS
 
 
-data Escape
-  = EscapeNormal
-  | EscapeUnicode !Int !Int
+data EscapeResult
+  = EscapeOk !Int ES.Chunk
```
avh4, line 265: ? might not have been necessary to combine these after all?

elm-format-lib/src/ElmFormat/Render/Box.hs
```diff
@@ -1944,15 +1944,29 @@ formatString elmVersion style s =
     stringBox quotes escaper =
       line $ row
           [ punc quotes
-          , literal $ escaper $ concatMap fix s
+          , literal $ escaper $ comonadic_ish_map fix s
```
avh4, line 1947: As you noted, we should look at doing this in the parse step so that the AST does not contain the CR's

elm-format-lib/src/Parse/Literal.hs
emmabastas: Error message added in 8dca014

elm-format-lib/src/Parse/Literal.hs
```diff
@@ -125,7 +97,7 @@ chr =
 
     toError e = newErrorUnknown ("Error parsing char: " ++ show e)
   in
-  do  s <- character toExpecation toError
+  do  s <- Parse.String.character newErrorUnknown newErrorUnknown
```
emmabastas: This is a silly mistake that does not compile, `[...] toExpectation toError` should not have been replaced. Fixed in ef5ef6a

elm-format-lib/src/ElmFormat/Render/Box.hs
emmabastas: b106094 reverts this commit.

356ce1b adds this feature to the parse step instead. When a CR is seen the string parsers look ahead and checks if there's an LF following. If so, it behaves as it would if encountering just an LF, but the pointer is incremented by two instead of one.

elm-format-lib/src/Parse/String.hs
emmabastas: This is adressed in 5a2d419.

#### Background.
TL;DR: I've changed my mind an now agree with Aarons comment, better to _not_ combine them after all.

The reason the Elm compiler has these two variants in the first place is because when encountering "normal escapes" (i.e `\[a-z]`), it can just include them as is in the JS output. Whereas `\u{....}` escapes needs to be transformed. In the original compiler code, `ES.Slice` is used for "normal escapes" and `ES.CodePoint` for `\u{....}` style.

https://github.com/avh4/elm-format/blob/778137761cd11f8a3010967d4fbd0ea8c945a0a1/elm-format-lib/src/Parse/String.hs#L79-L83

But in elm-format we normalize all escape sequences regardless so this distinction becomes less significant. The two branches are very similar now:

https://github.com/avh4/elm-format/blob/5a2d41983098bace966c082c638ff83668c87645/elm-format-lib/src/Parse/String.hs#L78-L82

An urge arises to generalize and have a single variant and branch for this which is indeed quite straightforward. But there is a value in having the code be similar to the compiler, which is why I agree with reverting to two variants again.

emmabastas:
Ping @avh4 PR is ready for round of reviews. In addition to responding to your comments I've also done some cleanup in ea8f4fe, 7879bff and dc6a7ed

(Are PRs supposed to feel this messy? 😅 At least the next PR ought to smaller in scope)

Issue #635, opened 2019-08-29
Elm format changes large integer literals


harrysarson:
The following elm code:

```elm
module Main exposing (toWrite)


largeInt: Int
largeInt=
    1000000000000000000000
```

is is changed by elm-format as such:

```diff
--- a/Main.elm
+++ b/Main.elm
@@ -3,4 +3,4 @@ module Main exposing (toWrite)
 
 toWrite : String
 toWrite =
-    String.fromInt 1000000000000000000000
+    String.fromInt 3875820019684212736
```

The code is malformed anyway due to https://github.com/elm/compiler/issues/1246 but this will become a problem as and when the elm compiler generates correct code for this. The underlying issue is the AST uses too small an integer datatype.

avh4:
Are you using the Windows binaries (which are compiled with 32-bit haskell),  or the Mac/Linux ones?

harrysarson:
Yes windows!

avh4:
Ah, okay, in that case I suspect it might be fixed if you were to compile elm-format from source with a 64-bit haskell installation.

I've only been releasing 32-bit windows binaries because it seemed to work and caused the least confusion, but I might try taking a look at building 64-bit also.

avh4:
Hmm... though it looks like this is reproducible on linux too 🤔

harrysarson:
Presumably the root cause is the same here and over at the elm compiler?

harrysarson:
The issue is that this number (`1000000000000000000000` or `0x36_35C9ADC5_DEA00000`) requires more than 64 bits to represent.

Elm format changes it to `3875820019684212736` or `0x35C9ADC5_DEA00000`, truncating it to the least significant 64 bits.

Whilst elm format may well not want to support such big numbers, it should at the least exit with an error when such a literal is encountered.

harrysarson:
What if elm-format did not try to encode literals as Int's in the AST but kept all literals as strings? This way there would be no encoding issues.

basile-henry:
`Integer` (big num) is pretty standard in Haskell. I think using it instead of `Int` is the simplest change. If you encode it as String you run the risk of allowing non integers to be represented.

avh4:
I think this is actually correct for now, as elm-format shows the user how elm-compiler will interpret the code.  When elm-compiler is changed, this should be revisited in elm-format.

harrysarson:
I would prefer to see elm-format raising a parsing issue rather than silently corrupting* source code files. 

[...]

PR #747, opened 2021-08-10
Use code from Elm compiler to parse numbers

emmabastas:
This PR focuses on migrating int and float literal parsing to use code from the Elm compiler.

### The commits in order of creation:

388e707 and 7b93d0f starts off by uncommenting and copying over declarations from the Elm compiler that will be needed for number parsing. a84ba10 fixes basic compile errors.

a73a37e, efbea12 and 1ae84b4 Includes the basic changes to the compilers number parsers to make them work with elm-format.

8450d4b Changes the Integer representation from `Int` to `Int64` to retain the behavior described in #635 (Weather or not this behavior is desirable or not might not be determined yet, but let's not change the behavior without thinking thoroughly about it first)

30b5241 Makes the number parsers work with Elm 10.17's `[n..m]`-style list literals.

And finally elm-format is made to use these parsers in e691d73


elm-format-lib/src/Parse/Number.hs
```diff
@@ -262,12 +263,12 @@ chompHexInt pos end =
 -- Return -2 if it has BAD digits
 
 {-# INLINE chompHex #-}
-chompHex :: Ptr Word8 -> Ptr Word8 -> (# Ptr Word8, Int #)
+chompHex :: (Integral i) => Ptr Word8 -> Ptr Word8 -> (# Ptr Word8, i #)
```
emmabastas, line 272: `chompHex` is used in `Parse.String` when parsing Unicode escapes. Changing from `Int` to `Int64` here would therefore spill over into `Parse.String`, `Elm.String` and `Reporting.Error.Syntax`. This is why this function has been given this more general type signature instead.

Not sure this is the right thing to do though. Other options could be just letting the type change spill over or use `fromIntegral` in to convert back to `Int` in `Parse.String`

elm-format-lib/src/Reporting/Error/Syntax.hs
```diff
@@ -490,9 +491,10 @@ data Escape
 
 data Number
   = NumberEnd
-  | NumberDot Int
+  | NumberDot Int64
   | NumberHexDigit
   | NumberNoLeadingZero
+  deriving Show
```
avh4, line 497: This is fine to keep either way, but I'm just curious if `Show` was added simply for debugging, or if it was necessary for things to compile?

elm-format-lib/src/Parse/Number.hs
```diff
@@ -129,7 +129,7 @@ chompInt !pos end !n =
       if isDecimalDigit word then
         chompInt (plusPtr pos 1) end (10 * n + fromIntegral (word - 0x30 {-0-}))
 
-      else if word == 0x2E {-.-} then
+      else if word == 0x2E {-.-} && not (isDot (plusPtr pos 1) end) then
```
avh4, line 132: 👍 And I think this is safe because previously going into `chompFraction` with a second `.` would always result in an error, and the only change now is that if there is a valid range syntax, that will be parsed as valid?

avh4 (approved):

elm-format-lib/src/Parse/Number.hs
emmabastas: Yes, that is correct.

This behavior of looking ahead for a second `.` also retains the same behavior as [Parse.Literals.decimals](https://github.com/avh4/elm-format/pull/747/commits/e691d73cda22d9a05cb3b53f7e28c00cfb9c5ab2#diff-aaaf99696880972a84fa197f3fb95b771e2ee24e7fbc6cc5e59ffe520c92c835L64) which was the function that parsed the decimal bit previously

elm-format-lib/src/Reporting/Error/Syntax.hs
emmabastas: This is actually needed for now. `show` is used chuck the errors onto a parsec error message.
Here's an example.

https://github.com/avh4/elm-format/blob/8bbada66239b8c0a7c217d6d80cac3db21e7bbff/elm-format-lib/src/Parse/Literal.hs#L23

I also vaguely recall some test helpers relying on `show` to display error messages.

avh4: Ah okay, so it's needed here but not in elm-compiler because we're using Strings for errors, instead of elm-compiler's customer error type

emmabastas: Yes,
if we ever get to the point where all of the parsers have migrated away from parsec then we'll be able to use Elm compiler-style errors all the way, and we wouldn't need `show` (at least not for this reason). Converting the errors to strings done when compiler-style parsers are used in parsec-style parsers, i.e at the border between the old and new code.

(spoiler: In the refactoring PR that I'm currently working on I'll be presenting an attempt at changing the parsec error type so that the Elm compiler's errors can be included without having to be converted to strings)

PR #748, opened 2021-08-17
Misc refactors to the Elm compiler's parser integration code

emmabastas:
Apart from small changes and improvements there is one notable refactors in this PR

### Refactor the parsec adapter's error type

---
elm-format doesn't report errors anyways, why care?
* Good error reporting can aid in debugging.
* We might want to report errors in some context (jsonAST?) at some point, at least let's not actively dismantle existing error reporting infrastructure.
* Great effort has been expended to give the Elm compiler good error reporting infrastructure, it would be a shame to not utilize it.
---

Parsec errors are essentially strings with some additional metadata. The Elm compiler on the other hand defines all possible parse errors as custom types inside `Reporting.Error.Syntax`. With this approach no information is lost when a parse error is propagated from the lowest level parser through the higher level ones. The higher level ones just construct higher level errors with the lower level errors as children.

393c7ee redefines parsecs error type. Instead of a single string with metadata it's now possible to create a tree structure, kind of like a generic/untyped version of the Elm compilers errors, it looks like this:
https://github.com/avh4/elm-format/blob/393c7eea66dd438a00bcdc66e2176eb8180f87d8/elm-format-lib/src/Parse/ParsecAdapter.hs#L348-L351
https://github.com/avh4/elm-format/blob/393c7eea66dd438a00bcdc66e2176eb8180f87d8/elm-format-lib/src/Parse/ParsecAdapter/Message.hs#L4-L7

9770e47  Integrates the parsec errors with the Elm compilers errors. Additional leaf variants are added to `Message` so that the error types from the compiler parsing code can be added as children to a parsec error without first needing to convert to a string. 

What the `Message` type looks like now: (note that `String`, `Char` and `Number` refer to custom error types)
https://github.com/avh4/elm-format/blob/9770e4744392fdb5441a1ebc89f654a7d86033b3/elm-format-lib/src/Reporting/Error/Syntax.hs#L105-L111


elm-format-lib/src/Parse/ParsecAdapter.hs
```diff
@@ -368,48 +345,51 @@ sourceColumn (SourcePos _ col) =
...
 
 
-newErrorMessage :: Message -> Row -> Col -> ParseError
-newErrorMessage msg row col
-    = ParseError row col [msg]
+errorPos :: ParseError -> SourcePos
+errorPos err =
+  case err of
+    Nil ->
+      error "An unexpeced error occured, this is likely a bug. Please report this issue at https://github.com/avh4/elm-format/issues"
```
emmabastas, line 313: `errorPos` crashes on empty error. I realize that it's possible to define `ParseError` to make impossible state impossible. Since Nil isn't exposed outside this module and since it's never used to construct an empty error I figured this was ok. That said, just let me know if you'd like this changed!

avh4:
This all looks good.  And some of these were pretty big cleanups -- thanks!

PR #769, opened 2022-05-05
Add macos aarch64 build support to Shakefile system

supermario:
As mentioned on Discord, here are the rough changes I made last weekend to generate an aarch64 build on my M1 Mac using `./build.sh --verbose --verbose -- dist`.

Feels a little hacky, so happy to take feedback and refine :)

### DIFF
--- a/Shakefiles/Extra.hs
+++ b/Shakefiles/Extra.hs
@@ -1,7 +1,8 @@
-module Shakefiles.Extra (phonyPrefix, forEach) where
+module Shakefiles.Extra (phonyPrefix, forEach, addMacOSGhcOptionFFI) where
 
 import Development.Shake
 import Data.List (stripPrefix)
+import Shakefiles.Platform (platform, Platform(..))
 
 
 phonyPrefix :: String -> (String -> Action ()) -> Rules ()
@@ -15,3 +16,12 @@ phonyPrefix prefix action =
 forEach :: Monad m => [a] -> (a -> m ()) -> m ()
 forEach list f =
     mapM_ f list
+
+
+--ghc-option required because of https://gitlab.haskell.org/ghc/ghc/-/issues/20592
+addMacOSGhcOptionFFI :: [String] -> [String]
+addMacOSGhcOptionFFI options =
+    case platform of
+        Mac -> ["--ghc-option=-I/Library/Developer/CommandLineTools/SDKs/MacOSX11.sdk/usr/include/ffi"] ++ options
+        MacArm64 -> ["--ghc-option=-I/Library/Developer/CommandLineTools/SDKs/MacOSX11.sdk/usr/include/ffi"] ++ options
+        _ -> options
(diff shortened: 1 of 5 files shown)


Shakefiles/Extra.hs
```diff
@@ -15,3 +16,12 @@ phonyPrefix prefix action =
...
 forEach list f =
     mapM_ f list
+
+
+--ghc-option required because of https://gitlab.haskell.org/ghc/ghc/-/issues/20592
+addMacOSGhcOptionFFI :: [String] -> [String]
+addMacOSGhcOptionFFI options =
+    case platform of
+        Mac -> ["--ghc-option=-I/Library/Developer/CommandLineTools/SDKs/MacOSX11.sdk/usr/include/ffi"] ++ options
+        MacArm64 -> ["--ghc-option=-I/Library/Developer/CommandLineTools/SDKs/MacOSX11.sdk/usr/include/ffi"] ++ options
```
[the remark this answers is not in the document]
avh4: The way to do that would be something like:

```haskell
StdoutTrim ffiFlags <- cmd "pkg-config" "--cflags" "libffi"
let ghcFfiFlags = "--ghc-option=" <> ffiFlags
```

(I think it's safe enough for now to assume the output won't contain stuff that needs to be escaped?  I'm not quite sure what'd be the easiest way to properly escape it.)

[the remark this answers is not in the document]
avh4: Actually, `Development.Shake.cmd` will escape the arguments as needed, so maybe it will be escaped properly?  (Depends on how `--ghc-option` parses any stuff after the `=` 🤷 )

Shakefiles/Extra.hs
```diff
@@ -15,3 +16,12 @@ phonyPrefix prefix action =
 forEach :: Monad m => [a] -> (a -> m ()) -> m ()
 forEach list f =
     mapM_ f list
+
+
+--ghc-option required because of https://gitlab.haskell.org/ghc/ghc/-/issues/20592
```
avh4, line 21: 💯 Thanks for the issue link!  And hopefully this will be fixed in ghc 9.4 🙏

stevenyap:
Following the PR by @supermario, I was able to successfully build `elm-format` on my Macbook Pro M1 Max.

Shakefiles/Extra.hs
avh4: FYI, this issue is supposedly fixed in ghc 9.2.3.  (No need to update this PR yet -- I'm doing a bit of other maintenance first.)

avh4: This is apparently still a problem with ghc 9.2.5.

But I ended up reverting the changes here to apply this in the Shakefiles, and instead it needs to be set in the outside environment.  (I'm not sure that's the ideal choice, but I'm gonna stick with it since it's working now for arm64 CI).  Here's the change I settled on: https://github.com/avh4/elm-format/pull/784/commits/4e5113a6a7014680302866a57da776afe0598e9d

avh4:
Thanks!  I'm going to include this in https://github.com/avh4/elm-format/pull/784, where I've made a few small changes to what you provided here.

Issue #755, opened 2021-12-08
process multiple files in parallel


avh4:
Processing of each file is independent.  Would it be faster overall if they were allowed to run in parallel?  I'd expect that elm-format is mostly disk-bound, but maybe that's not the case with SSD drives or with caching in modern systems?

If it can be faster, we'd want to be careful to use a haskell library that would appropriately manage resources and not start additional threads if CPU usage or memory usage is already high.

lydell:
This would be great, I think!

At work we have 726 Elm files. Letting one `elm-format --validate` check them all takes 15 seconds, while spreading the files out evenly on multiple `elm-format --validate` calls in parallel takes down to 4 seconds.

I tried a bunch of different numbers of parallel `elm-format --validate`, up to 12 which is the number of cores on my laptop.

- 1: 14.37 s
- 2: 8.26 s
- 4: 4.88 s
- 6: 4.31 s
- 8: 3.96 s
- 10: 3.93 s
- 12: 3.85 s

Conclusions:

- There looks to be a speedup from parallelization here!
- Even using just 2 cores gives a speedup, so even GitHub Actions would benefit.

Issue #183, opened 2016-05-20
Improve performance


avh4:
Currently formatting 400 files takes ~2 seconds on a 2015 MBP.  Can that time be cut in half?

Janiczek:
Are the files used available anywhere? (Maybe there's an outlier.) Also, how does it happen that you format 400 files at once?

avh4:
Nope, the files aren't available.  So maybe the first task is to assemble a test suite.  (Maybe bulk formatting the existing test suite would do?)

On the set of files I have, there is not single file that seems slow.

This results from bulk-formatting all files in a directory.  But also is when running `elm-format --validate` either on CI or in scripts.  I think this is also a valid use case for teams that want to run elm-format as a commit hook instead of on-save in their editors.

jpierson:
As indicated in https://github.com/avh4/elm-format/issues/408#issuecomment-478240826 I have a particular case with a somewhat large Elm file that is taking elm-format around 13 seconds to format.

I've tested this from the terminal using the following command:

```powershell
> measure-command { npx elm-format .\src\Pages\PersonalInfoPage.elm --yes }
```
This yields the following result:

```txt
...
TotalSeconds      : 13.8024885
...
```

Running the same command but for a different Elm source file which is only *55 lines* long takes *767 milliseconds*. While this is still slow it's still in the realm of feasibility in terms of an editor auto save experience.

What I'm wondering is whether there has been a performance regression Windows specifically or something along those lines with recent versions? I say this because I remember running Elm format in the past with no problems so either this is the first time I'm working with Elm source files this large or the performance of elm-format has regressed since earlier versions I've used in the past. Note that Evan has been recently advising others to starting with a file instead of splitting up files prematurely so that it's possible that other developers (perhaps only a subset on specific platforms) will began experiencing these issues more.

folkertdev:
This is the profiling result of elm-format on elm-spa-example:

```
	   elm-format +RTS -p -RTS src/ src/Api/Endpoint.elm src/Article/Body.elm src/Article/Comment.elm src/Article/Feed.elm src/Article/Slug.elm src/Article/Tag.elm src/Page/Article.elm src/Page/Article/Editor.elm src/Page/Blank.elm src/Page/Home.elm src/Page/Login.elm src/Page/NotFound.elm src/Page/Profile.elm src/Page/Register.elm src/Page/Settings.elm --yes

	total time  =        2.85 secs   (2847 ticks @ 1000 us, 1 processor)
[...]

PR #770, opened 2022-08-28
Process multiple files in parallel

lydell:
Fixes https://github.com/avh4/elm-format/issues/755. Related to https://github.com/avh4/elm-format/issues/183.

On my computer, it’s roughly 2.5 times faster! :tada:

| command | main | this PR |
| - | - | - |
| elm-format --validate . | 7.7 s | 2.9 s |
| elm-format --yes . >/dev/null | 7.7 s | 2.9 s |
| elm-format --yes . | 7.7 s | 3.3 s |

`main` refers to commit 15578927a7fd327abd9c619031243c0abe96c57b.

Tested on a code base of this size:

```
❯ tokei --type Elm
===============================================================================
 Language            Files        Lines         Code     Comments       Blanks
===============================================================================
 Elm                   745       221612       178984         5131        37497
```

Computer specs:

- macOS Monterey
- CPU: 2,6 GHz 6-Core Intel Core i7
- RAM: 16 GB 2400 MHz DDR4

Terminal: `elm-format --yes .` (without `>/dev/null`) depends on your Terminal. With the default Apple Terminal, I got 3.3 seconds as shown in the table above. With iTerm2, I got 5.5 seconds (and 9.7 seconds for `main`).

Notes:

- This PR seems to work, but I haven’t tested all aspects extensively.
- I have only tested on macOS, not Linux or Windows.
- I’m a Haskell beginner, so I’ve probably done some weird things. I even know some of them: I generate JSON for the list of validation objects a bit manually, and I turn errors into JSON earlier than feels logical to avoid a “deriving NFData hell” that I couldn’t figure out. I took all shortcuts I could just to win over the compiler :)
- It feels like I completely changed how `InfoFormatter` works. Sorry if that doesn’t fit with your vision for it!
- I use @supermario’s hack to avoid garbled printouts: https://github.com/supermario/elm-tooling-compiler/blob/0f63c27e0f334ca680a568034953f9b3d16ba3db/ext-common/Ext/Common.hs#L82-L94
- I didn’t spend much time thinking about code style. It’s probably all over the place and not aligned with the rest of the code base!
- I use [pooled-io](https://hackage.haskell.org/package/pooled-io-0.0.2.2) which is on version 0.0.2.2 – that’s a very low number! I picked that dependency because it was the only one I understood how to use. It is only used to implement `World.mapMConcurrently` (a one-liner) though, so it might be easy to replace by someone better at Haskell.


avh4:
Awesome, thanks!!

- Linux: I'll be able to test that
- Windows: I'll make sure to have someone check a release candidate build before the next release (or I guess I need to make github CI save the windows artifacts for PRs)
- pooled-io 0.0.2.2: low version number doesn't bother me too much, as long as the library is pretty small and self-contained.  Also in this case, I did look at pooled-io's source code, and it looks pretty straightforward as a wrapper over more common Haskell concurrency APIs, so this seems totally fine to me
- code style: I haven't really settled on a consistent one anyway 😅.  Everything I saw so far looks good for code style.
- InfoFormatter: I've been trying to get rid of it for a while, so I'm pretty much fine with anything here

Apologies, I'll probably be a little bit slow on merging this since I want to re-read the InfoFormatter changes a couple times, and make sure I personally understand how Haskell's laziness works w/r to the new code.  (I trust that doesn't really break anything, both from a quick skim of the code, and from the testing you've done; but I want to put a bit higher of a bar on myself to understand what the runtime behavior actually is.)

Here are a couple things I'll want to take a look at before I merge.  Feel free to take a stab at them if you want, or no worries if you don't, I should have a chance to get to them in the next couple weeks.

- I'm thinking using the print lock, it should be possible now to have the forked jobs just print their output, and then only have to pass back a Bool to the parent, and that maybe also would restore the streaming output of the json
- I'd like to get rid of the unsafePerformIO hack, which I think I'd do by making a new monad stack for the "real" `World` instance, which would be something like `type RealWorld a = ReaderT (MVar ()) IO`, make a `MonadIO RealWorld` instance (maybe?  I think this might make it less verbose to refactor the World instance?), and then refactor `instance World IO` to be `instance World RealWorld` (and the top-level entrypoint then will create the printLock and use `runReaderT printLock`
- (probably won't look at now for this PR, but might someday in the future) rewrite TransformFiles to use [conduit](https://hackage.haskell.org/package/conduit), and then use [conduit-concurrent-map](https://hackage.haskell.org/package/conduit-concurrent-map-0.1.3/docs/Data-Conduit-ConcurrentMap.html) instead of pooled-io.

avh4:
Testing version for MacOS arm64 is available here: https://github.com/avh4/elm-format/actions/runs/4141690402

avh4:
Just an additional benchmark on a Mac M1 using `--validate` on ~150k LOC:

(using https://github.com/sharkdp/hyperfine `hyperfine --warmup 3 --shell=none`)

| Version | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| **darwin-arm64** |
| this PR | 0.824 ± 0.006 | 0.815 | 0.832 | 1.00 |
| `main` | 2.776 ± 0.015 | 2.758 | 2.805 | 3.37 ± 0.03 |
| **darwin-x86 via Rosetta** |
| this PR | 1.058 ± 0.012 | 1.042 | 1.085 | 1.28 ± 0.02 |
| `main` | 9.603 ± 0.096 | 9.475 | 9.770 | 11.65 ± 0.14 |

avh4:
> I'm thinking using the print lock, it should be possible now to have the forked jobs just print their output, and then only have to pass back a Bool to the parent, and that maybe also would restore the streaming output of the json

I took a brief look at doing this now, but it doesn't seem worth the time atm.  Doing it in a clean way seems like it would require a way of coordinating an additional thread, which would fit better with the possible future work of using `conduit`.

avh4:
Split off https://github.com/avh4/elm-format/issues/788 for possible future refactoring.

avh4:
Merged via https://github.com/avh4/elm-format/pull/789

Thank you!!

Issue #765, opened 2022-04-14
No binaries are available for your platform: linux-arm64


os6sense:
Attempting to install and build on a linux based container on an M1 using 0.8.5:

```
       → error /app/node_modules/elm-format: Command failed.
       → Exit code: 1
       → Command: binwrap-install
       → Arguments: 
       → Directory: /app/node_modules/elm-format
       → Output:
       → /app/node_modules/binwrap/install.js:18
       →     throw new Error("No binaries are available for your platform: " + buildId);
       →     ^
       → 
       → Error: No binaries are available for your platform: linux-arm64
       →     at install (/app/node_modules/binwrap/install.js:18:11)
       →     at Object.install (/app/node_modules/binwrap/index.js:14:14)
       →     at Object.<anonymous> (/app/node_modules/binwrap/bin/binwrap-install:18:9)
       →     at Module._compile (node:internal/modules/cjs/loader:1103:14)
       →     at Object.Module._extensions..js (node:internal/modules/cjs/loader:1157:10)
       →     at Module.load (node:internal/modules/cjs/loader:981:32)
       →     at Function.Module._load (node:internal/modules/cjs/loader:822:12)
       →     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
       →     at node:internal/main/run_main_module:17:47
       → info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

This is a slightly different scenario over running this on darwin-arm64 since here I'm attempting to build and run in a container.

Zeneixe:
I have the same problem

avh4:
I'm trying to get linux-arm64 binaries build for the next release.  Can anyone test running the aarch64 binary here: https://elm-format.avh4.net/manual/0.8.5-424-g9e071c6f/ and see if it works in the Linux ARM64 environment that they need it for?

jsynacek:
> Can anyone test running the aarch64 binary here:

Seems to be working well for me.

```
$ uname -a
Linux 7bed0a212d7c 5.15.49-linuxkit #1 SMP PREEMPT Tue Sep 13 07:51:32 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux
```

avh4:
Fixed in https://github.com/avh4/elm-format/pull/798  The next release (0.8.6) will include official linux-aarch64 static binaries.

Thanks @jsynacek for testing!

jsynacek:
[...]

Issue #714, opened 2020-10-18
Build linux-aarch64 (arm64) release binaries


avh4:
This should be done in a reproducible way, preferably by being able to build locally using docker and buildx support for a locally-running arm container win docker-supported qemu.

Starting point: https://www.docker.com/blog/multi-arch-images/
https://docs.docker.com/buildx/working-with-buildx/
https://tech.smartling.com/building-multi-architecture-docker-images-on-arm-64-c3e6f8d78e1c

`docker build ...` becomes `docker buildx build --platform linux/arm64 ...`

Related:
- possible dockerfile for aarch64 ghc https://github.com/haskell/docker-haskell/issues/1
- discussion of aarch64 alpine ghc https://gitlab.haskell.org/ghc/ghc/-/issues/18039

Blockers:
- [x] upgrade elm-format to ghc >= 8.10.1 (introduces aarch64 support) https://github.com/avh4/elm-format/pull/724
- [x] ghcup support for aarch64 https://gitlab.haskell.org/haskell/ghcup-hs/-/issues/5
- [x] cabal-install aarch64 binary https://github.com/haskell/cabal/issues/6715

avh4:
See https://github.com/mpizenberg/elm-test-rs/pull/91/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR44-R54 regarding github CI runners

Zeneixe:
Any chance this platform's support will be added? @avh4

avh4:
I might be getting some new hardware that'll be able to build aarch64 binaries soon.

In the meantime, the issue is that Github Actions CI (or any other free-for-open-source-projects CI that I've found so far) doesn't provide any aarch64 or arm64 runners yet afaik.  If anyone has any info about that, please share it!

Zeneixe:
@avh4 I noticed some people suggest using this:

https://github.com/uraimo/run-on-arch-action

https://github.com/marketplace/actions/run-on-architecture

avh4:
> https://github.com/uraimo/run-on-arch-action

Ooh, that's great!  Thanks!

ParetoOptimalDev:
Hi, I use NixOS to successfully and simply build `elm-format` for aarch64 like:

```
nix-build -A pkgsCross.aarch64-multiplatform-musl.pkgsStatic.elmPackages.elm-format
```

[...]

PR #777, opened 2022-09-28
Allow building on Linux ARM64

tfausak:
This would address #714 and #765. It's similar to #769. 

These changes allow `elm-format` to be built on an ARM64 (AArch64) Linux machine. I have tested them with AMD64 (x86-64) as well, which continues to work as expected.


ismaelga:
Bumped into the same issue. This PR seems legit.
Thanks! Hope it gets merged

avh4:
Huh, github seems confused about the list of commits after I fixed the merge conflicts here.  :shrug: 

Merging.

avh4:
Thanks!  This is on `main` branch now.  I haven't set up linux aarch64 CI yet, so lmk if there's still any problems.

jsynacek:
Awesome, thanks a lot! Would you mind releasing a new version so that the builds show in the release and also in the npm archives?

avh4:
The release is tracked here: https://github.com/avh4/elm-format/milestone/31

Basically I have to do a test run of the new npm publishing process, then get linux-aarch64 binaries to build on CI, then take a look at building musl binaries, and then do the release. 🤞

Issue #779, opened 2022-11-24
Critical severity vulnerability - json-schema (via binwrap)


sporto:
This package uses `binwrap` 0.2.3

Which depends on `request` 2.88.0
https://github.com/avh4/binwrap/blob/main/package.json#L42

Which depends on form-data 2.3.3
https://github.com/request/request/blob/master/package.json#L36

Which depends on json-schema 0.2.3

json-schema 0.2.3 has a critical vulnerability:
https://github.com/advisories/GHSA-896r-f27r-55mw

It would be great if this dependecy could be changed.

avh4:
The next release (0.8.6) will have no external npm dependencies, thanks to https://github.com/avh4/elm-format/pull/781.

PR #781, opened 2023-01-06
Free the npm package from third party dependencies

lydell:
This is:

- Based on https://github.com/elm/compiler/pull/2287
- Backwards compatible (same module interface, same environment variables)
- Free from third party dependencies 🎉 

I’ve documented a bit how to publish. If you want to automate that, feel free to do so. I won’t be doing any such things in this PR though.

Read the PR description as well as the changes in PUBLISHING.md in https://github.com/elm/compiler/pull/2287 for full details on how it works, the pros, the (minor) cons, and the full compatibility story.

Fixes https://github.com/avh4/elm-format/issues/779. There might be more issues to go through and see if they are still relevant: https://github.com/avh4/elm-format/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+binwrap


package/npm/package.json
```diff
@@ -2,13 +2,11 @@
   "name": "elm-format",
   "version": "0.8.5",
   "description": "Install elm-format",
-  "preferGlobal": true,
```
lydell, line 5: `preferGlobal` has been deprecated for a long time, and is even removed in newer npm versions.

avh4:
FYI, I'm gonna take a look at this after I get arm64 builds working.  (https://github.com/avh4/elm-format/milestone/31 should be in the correct order now.)

package/npm/package.json
```diff
@@ -1,14 +1,12 @@
 {
   "name": "elm-format",
-  "version": "0.8.5",
+  "version": "0.8.6-0",
```
lydell, line 3: We learned in https://github.com/mpizenberg/elm-test-rs/pull/121#issuecomment-1374454965 that it usually takes a few tries publishing everything correctly the first time, so I went ahead and added `-0` to everything. Once it’s working, we can release the top package as just `0.8.6`. Read about versioning of the sub packages here: https://github.com/elm/compiler/pull/2287/files#diff-cd34b117c7124e742fd6eb9ffd8f6f9f5c125402a66e9e62c906785b940f7681R68-R85

package/npm/packages/elm-format-darwin-x64/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@avh4/elm-format-darwin-x64",
+  "version": "0.8.6-0",
+  "description": "The macOS 64-bit binary for elm-format.",
+  "repository": "https://github.com/avh4/elm-format",
+  "license": "BSD-3-Clause",
+  "os": [ "darwin" ],
+  "cpu": [ "x64", "arm64" ]
+}
```
avh4, line 9: :thinking: does npm require this file to (and the README) to be in the repo somewhere?  I'm wondering if we could maybe just generate these folders when they're actually needed for publishing?

package/npm/packages/elm-format-linux-x64/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@avh4/elm-format-linux-x64",
+  "version": "0.8.6-0",
+  "description": "The Linux 64-bit binary for elm-format.",
+  "repository": "https://github.com/avh4/elm-format",
+  "license": "BSD-3-Clause",
+  "os": [ "linux" ],
+  "cpu": [ "x64" ]
```
avh4, line 8: Do you happen to know where the list of possible cpu values is?  (Or for the short-term, do you know what this value will be for linux arm64 and for MacOS arm64?)

package/npm/packages/elm-format-darwin-x64/package.json
lydell: We can generate them if you want. I just found it the easiest to have actual files so it’s possible to do manual fixups.

Not sure yet if _I_ will have time to do generation stuff though. A potential approach here is if you’d like to do that later in another PR.

package/npm/packages/elm-format-linux-x64/package.json
lydell: The package.json docs weirdly don’t mention what the values are: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#cpu

But this is the list: https://nodejs.org/api/process.html#processarch

> 'arm', 'arm64', 'ia32', 'mips','mipsel', 'ppc', 'ppc64', 's390', 's390x', and 'x64'

package/npm/packages/elm-format-darwin-x64/package.json
avh4: Sounds good!  I'm gonna take a look at generation on a new branch, then.

PUBLISHING.md
```diff
@@ -59,6 +59,22 @@ Then `cd elm-tooling`, push the resulting branch, and make a PR to <https://gith
...
+1. Copy the appropriate binary to `./elm-format`. For Windows: `./elm-format.exe`
+1. Double-check that you put the right binary in the right package: `file elm-format`
+1. Double-check that the file is executable: `ls -l elm-format`
+1. In `package.json` of the binary package, bump the version for example to `"0.8.6"`.
+1. In `package.json` of the main npm package, update `"optionalDependencies"` to point to the bumped version. For example: `"@avh4/elm-format-darwin-x64": "0.8.6"`.
+
+   Note: Pin the versions of the binary packages _exactly_ – no version ranges. This means that installing `elm-format@0.8.6` installs the exact same bytes in two years as today.
+1. Publish the package: `npm publish --access=public`
+
+   `--access=public` is needed because scoped packages are private by default.
```
avh4, line 74: And just to double-check that I'm understanding how this works, the binary will actually be published into the npmjs repo (since there's no explicit manifest given, all the files in the local directory get published?), so it's now actually npm that's downloading the binary?

lydell: Yes, the binary will end up in the tarball for the package and be stored in the npmjs registry, and npm will download the package as any other. An npm package is just a folder with a valid package.json and random files that npm doesn’t really care about.

Here’s an example: https://unpkg.com/browse/@lydell/elm_darwin_arm64@0.19.1-2/

If you `cd` into a folder with a `package.json` and run `npm publish`, here are some rules:

1. The contents of that folder is packed into a tarball and uploaded to the npm registry.
2. You can use an `.npmignore` file to exclude files, `.gitignore`-style.
3. You can use the `"files"` field in package.json to instead _only_ include the stuff listed there.
4. package.json is always included no matter what you do.
5. Some other files are always included as well, if they exist, such as README.md and LICENSE.

https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files

I didn’t include the `"files"` field in the sub packages because:

1. I based much of the approach on esbuild, and they don’t have it: https://github.com/evanw/esbuild/blob/e345b13687bc3ac86f18f4a266a162653544ad31/npm/%40esbuild/darwin-arm64/package.json
2. It felt unlikely that there will be other random files in those folders.
3. It saves some bytes? 🙃 

But you could add `"files": ["elm-format"]` (`"files": ["elm-format.exe"]` for Windows) if you want, it doesn’t hurt. (package.json and README.md are automatically included always.)

avh4: Awesome, thanks for the extra details and the unpkg.com/browse trick!

package/npm/package.json
avh4: Ah, okay, so if I'm understanding correctly:

- binary packages will always be name=`@avh4/elm-format` version=`$OS-$ARCH-$MAJ.$MIN.$PAT-$BINVER`
- for pre-release versions, the base package will be name=`elm-format` version=`$MAJ.$MIN.$PAT-$PREVER`
- once finalized, the base package will be name=`elm-format` version=`$MAJ.$MIN.$PAT`

avh4:
FYI, I started a new PR for automating the package generation https://github.com/avh4/elm-format/pull/795, which will include this PR when it's merged.

Thanks for the help!

package/npm/package.json
lydell: I would put it like this:

- The base package follows regular npm semver
- The binary packages don’t matter at all how you version, except for your own sanity
- The binary packages have the name `@avh4/elm-format-$OS-$ARCH`

avh4:
Merged via https://github.com/avh4/elm-format/pull/795

Thank you!!

I did a trial run published as `@avh4/elm-format-test@rc`, which seems to work on all platforms.  I also added a github workflow that checks installing the published packages: https://github.com/avh4/elm-format/actions/runs/4209086779

(and FYI, I'll do one last trial run as `elm-format@0.8.6-rc.1` once 0.8.6 is ready)

PR #803, opened 2023-03-08
Use hpack

avh4:
Fixes https://github.com/avh4/elm-format/issues/800


avh4:
Ah, "Build Release" CI builds are broken until #799 is merged.

PR #811, opened 2023-03-31
Remove unused packages

8n8:
Tidied up a package.yaml file by removing some unused dependencies.

### DIFF
--- a/elm-format-lib/package.yaml
+++ b/elm-format-lib/package.yaml
@@ -23,24 +23,18 @@ default-extensions:
 ghc-options:
   - -O2
   - -Wall
+  - -Wunused-packages
   - -Wno-name-shadowing
 
 dependencies:
   aeson: ">= 2.1.0.0 && < 2.2"
-  ansi-terminal: ">= 0.11 && < 0.12"
-  ansi-wl-pprint: ">= 0.6.9 && < 0.7"
-  array: ">= 0.5.4.0 && < 0.6"
   base: ">= 4.15.0.0 && < 5"
   bimap: ">= 0.4.0 && < 0.6"
   binary: ">= 0.8.9.0 && < 0.9"
   bytestring: ">= 0.11.3.0 && < 0.12"
   containers: ">= 0.6.5.1 && < 0.7"
-  directory: ">= 1.3.6.2 && < 2"
-  filepath: ">= 1.4.2.1 && < 2"
-  ghc-prim: ">= 0.5.2"
   mtl: ">= 2.2.2 && < 3"
   optparse-applicative: ">= 0.17.0.0 && < 0.18"
-  process: ">= 1.6.13.2 && < 2"
   relude: ">= 1.1.0.0 && < 1.2"
   text: ">= 2.0 && < 3"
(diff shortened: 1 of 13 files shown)


avh4:
Thanks!  I'll get to this soon (and before the 0.8.7 release).

I'll also add `-Wunused-packages` to all the other package.yaml's in the project -- feel free to add that yourself if you get to it before I do.

avh4:
I think I'm also gonna upgrade this to `-Werror=unused-packages` (which can be downgraded again in `cabal.project.local` if desired to speed up local development, but will help me remember to actually notice the warning when committing)

elm-format-lib/package.yaml
```diff
@@ -23,24 +23,18 @@ default-extensions:
...
-  ansi-wl-pprint: ">= 0.6.9 && < 0.7"
-  array: ">= 0.5.4.0 && < 0.6"
   base: ">= 4.15.0.0 && < 5"
   bimap: ">= 0.4.0 && < 0.6"
   binary: ">= 0.8.9.0 && < 0.9"
   bytestring: ">= 0.11.3.0 && < 0.12"
   containers: ">= 0.6.5.1 && < 0.7"
-  directory: ">= 1.3.6.2 && < 2"
-  filepath: ">= 1.4.2.1 && < 2"
-  ghc-prim: ">= 0.5.2"
```
avh4, line 40: @8n8 Do you remember why you had to re-add ghc-prim in your initial PR?  It made me remove it again when I made the warning into an error, and it compiled fine for me locally.  I guess I'll see if CI fails.

8n8: Yes, I removed it when I saw the warning, but then when I built again from scratch it failed. I think I had to do `git clean -f` to remove the build caches to make it fail.

8n8: If you replace `import GHC.Prim` with `import GHC.Exts` in `elm-format-lib/src/Data/Utf8.hs` it fixes the issue.

The [docs](https://hackage.haskell.org/package/ghc-prim) for GHC.Prim say:

> This package contains the primitive types and operations supplied by GHC. It is an internal package, only for the use of GHC developers. GHC users should not use it! If you do use it then expect breaking changes at any time without warning. You should prefer to import GHC.Exts from the base package instead.

avh4: Ah, great!  I'll make the change to `GHC.Exts`.  I also reported [#23212: -Werror=unused-packages cannot be satisfied for ghc-prim · Issues · Glasgow Haskell Compiler / GHC · GitLab](https://gitlab.haskell.org/ghc/ghc/-/issues/23212).

avh4:
Thanks for the cleanup!

elm-format-lib/package.yaml
avh4: oh, also `dev/build.sh clean && dev/build.sh` was enough to trigger the original problem without having to re-clone.

8n8: Oh right. That's easier, thanks.

Issue #301, opened 2017-03-05
Use new elm-compiler parser


avh4:
elm-compiler has a new parser, as of 2017-12.  elm-format should switch to the new parser, as it is supposedly faster, and also gives much better error messages in most cases.

avh4:
Discussed this w/ Evan, and the plan here should be to copy over `Primitvies.hs`, and then implement all required Parsec primitives on top of the new Primitives, then drop parsec, and then the remainder of the parser can be incrementally rewritten to use the new primitives directly.

PR #812, opened 2023-04-01
Removed unused code in elm-format-lib

8n8:
Removed some unused code.

### DIFF
--- a/elm-format-lib/src/Parse/Primitives.hs
+++ b/elm-format-lib/src/Parse/Primitives.hs
@@ -27,8 +27,7 @@ import qualified Data.ByteString.Internal as B
 import Data.Word (Word8, Word16)
 import Foreign.Ptr (Ptr, plusPtr)
 import Foreign.Storable (peek)
-import Foreign.ForeignPtr (ForeignPtr, touchForeignPtr)
-import Foreign.ForeignPtr.Unsafe (unsafeForeignPtrToPtr)
+import Foreign.ForeignPtr (ForeignPtr)
 
 import qualified Reporting.Annotation as A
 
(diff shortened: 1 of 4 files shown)


elm-format-lib/src/Parse/Primitives.hs
```diff
@@ -202,36 +201,6 @@ instance Monad (Parser x) where
...
---  B.accursedUnutterablePerformIO $
---    let
---      toOk' = toOk toBadEnd
---      !pos = plusPtr (unsafeForeignPtrToPtr fptr) offset
---      !end = plusPtr pos length
---      !result = parser (State fptr pos end 0 1 1) toOk' toOk' toErr toErr
---    in
---    do  touchForeignPtr fptr
---        return result
-
```
avh4, line 220: This will eventually be needed; ideally https://github.com/avh4/elm-format/blob/main/elm-format-lib/src/ElmFormat/Parse.hs#L22 will change to take ByteString instead of Text, and https://github.com/avh4/elm-format/blob/aded67fb5db42afd17bfa372eab23f8a01cb635d/elm-format-lib/src/Parse/Parse.hs will change to take ByteString instead of String.

Sorry things are a bit of a mess atm.

elm-format-lib/src/Reporting/Annotation.hs
```diff
@@ -58,11 +57,6 @@ instance Coapplicative Located where
     {-# INLINE extract #-}
 
 
-traverse :: (Functor f) => (a -> f b) -> Located a -> f (Located b)
-traverse func (At region value) =
-  At region <$> func value
```
avh4, line 63: I'm a little hesitant to remove, as it might be useful for some upcoming features.  Though maybe this should be moved to a `Traversable` instance?

elm-format-lib/src/Parse/Primitives.hs
avh4: Specifically, https://github.com/avh4/elm-format/issues/301 is the blocking issue for the above mentioned ideal changes.

PR #815, opened 2023-04-01
Backfill test for using `=>` as an operator in Elm <= 0.18

avh4:
As follow-up to https://github.com/avh4/elm-format/pull/809, this adds a test to ensure compatibility with Elm <= 0.18


avh4:
@kutyel FYI adding the test we discussed on Slack

PR #832, opened 2025-06-15
Fix arrow indentation in case patterns with comments

### LINKED ISSUE #824: Formatting produces invalid code when comment is in pattern of a case expression
When introducing a comment in a pattern in a case expression, `elm-format` will reformat it like the following:
```elm
value =
    case x of
        A
            -- comment
            ()
        ->
            1

        B
        -- comment
        ->
            1
```

Here, both branches are considered to be a syntax error. The `->` needs to be indented by at least one space (or move to after the `()` in the first pattern).
(more complete example: https://ellie-app.com/rHrcvrMP88Ha1)

sandydoo:
Fixes #824.

The example from the issue would be formatted as follows. Case A looks a little funky to me, but I couldn't find any existing examples of a multi-line pattern match with comments and associated data. The arrows stay in one column though, so that's good.

```elm
value =
    case x of
        -- A
        A
            -- B
            ()
            ->
                1

        -- C
        B
        -- D
            ->
                1
```

### DIFF
--- a/elm-format-lib/src/ElmFormat/Render/Box.hs
+++ b/elm-format-lib/src/ElmFormat/Render/Box.hs
@@ -1527,10 +1527,10 @@ formatExpression elmVersion importInfo aexpr =
                                    , indent body'
                                    ]
                         (_, _, pat', body') ->
-                            stack1 $
+                            stack1
                               [ pat'
-                              , line $ keyword "->"
-                              , indent body'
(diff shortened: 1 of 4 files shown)


avh4:
Perfect, thank you!  I think I can do the release on Weds.

avh4:
@sandydoo FYI I published the 0.8.8 release on github and on npm

PR #841, opened 2026-06-07
Relax upper bound on optparse-applicative

turboMaCk:
This simply relaxes upper bound on optparse-applicative. No additional changes needed. This makes it possible to build elm-format in more stackage snapshots and removes need for jailbreak in nixpkgs

https://github.com/NixOS/nixpkgs/pull/529072

As a side note - I would consider dropping hpack - it has many problems, is considered legacy and doesn't generate cabal files for this project correctly.

### DIFF
--- a/dev/build.sh
+++ b/dev/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 (cd Shakefile && hpack)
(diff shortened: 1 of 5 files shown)


elm-format.cabal
```diff
@@ -70,7 +70,6 @@ test-suite elm-format-tests
       ElmFormat.CliFlags
       ElmFormat.Messages
       ElmFormat.Version
-      Build_elm_format
```
turboMaCk, line 73: hpack decided to remove this. It seems to be not needed.

dev/build.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
turboMaCk, line 1: this makes build script more portable

